PR #10, opened 2020-06-07
Unregister shortcut before setting a new one

ywolff:
I encountered the following bug using `RecorderCocoa`:
- the user configures a custom shortcut which is already used by the system (`CMD+A` for instance)
- this custom shortcut overwrites the system shortcut (desired behavior)
- the user changes the shortcut with a new one, by clicking into the `RecorderCocoa` input
- then, this new shortcut works, but the default behavior of the previous shortcut (`CMD+A`) does not work anymore, until the app is killed.

This bug does not occur when the user clicks on the "X" button before setting the new shortcut. It appears that in this case, the method `userDefaultsRemove` manages to properly remove the current shortcut. In this PR, I'm adding a call to this method before setting the new shortcut, even when the user does not click on that "X" button.

PS: thanks a lot for this awesome package!

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -197,8 +197,12 @@ public enum KeyboardShortcuts {
 			return
 		}
 
-		UserDefaults.standard.set(encoded, forKey: userDefaultsKey(for: name))
+		if let oldShortcut = userDefaultsGet(name: name) {
+			unregister(oldShortcut)
+		}
+
 		register(shortcut)
+		UserDefaults.standard.set(encoded, forKey: userDefaultsKey(for: name))
 		userDefaultsDidChange(name: name)
 	}
 


sindresorhus:
> this custom shortcut overwrites the system shortcut (desired behavior)

How do you achieve this? This check should make sure such shortcut is never set: https://github.com/sindresorhus/KeyboardShortcuts/blob/d07dee7479759c64bbee2200e0150fae418238c2/Sources/KeyboardShortcuts/RecorderCocoa.swift#L190-L203

To conform this behavior, I set a breakpoint at: https://github.com/sindresorhus/KeyboardShortcuts/blob/d07dee7479759c64bbee2200e0150fae418238c2/Sources/KeyboardShortcuts/RecorderCocoa.swift#L208

And ran the example app, tried Cmd+A (many times), and I could never get the breakpoint to hit. What am I missing?

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -205,6 +205,7 @@ extension KeyboardShortcuts {
 				self.stringValue = "\(shortcut)"
 				self.showsCancelButton = true
 
+				userDefaultsRemove(name: self.shortcutName)
 				userDefaultsSet(name: self.shortcutName, shortcut: shortcut)
```
sindresorhus, line 209: I think it would be better for `userDefaultsSet` to handle unregistering any existing shortcut.

ywolff:
Thanks fr your answer.
As far as I understand, the `CMD+A` shortcut is not a system shortcut (indeed, you can't find it in “System Preferences › Keyboard › Shortcuts“). I suppose that it's not the `shortcut.isTakenBySystem` property which prevents you to choose this shortcut, but the `shortcut.takenByMainMenu`. Indeed, in your `KeyboardShortcutsExample`, this shortcut is used by the Main Menu:
image
In the app I'm developing, I have deleted this default Main Menu shortcut from the Main storyboard, that's why I can override this shortcut.

Sources/KeyboardShortcuts/RecorderCocoa.swift
ywolff: I've updated my PR, by calling the `userDefaultsRemove` method inside `userDefaultsSet`. Is it the way you had in mind?

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -193,6 +193,8 @@ public enum KeyboardShortcuts {
 	}
 
 	static func userDefaultsSet(name: Name, shortcut: Shortcut) {
+		userDefaultsRemove(name: name)
```
sindresorhus, line 196: It should be placed after line 200.

And thinking more about this, I think it's enough to just call `unregister(shortcut)`, since `UserDefaults.standard.removeObject` (in the `userDefaultsRemove` method) is moot since we'll set the key right away afterwards.

sindresorhus: So I would make it:

```swift
if let oldShortcut = userDefaultsGet(name: name) {
	unregister(oldShortcut)
}

register(shortcut)
UserDefaults.standard.set(encoded, forKey: userDefaultsKey(for: name))
userDefaultsDidChange(name: name)
```

Can you do that and test that it still works fine in your app?

ywolff: I've just updated, and that works in my app, thanks.

sindresorhus:
Thanks :)

https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/v0.2.2

PR #12, opened 2020-06-15
Add optional change callback to `Recorder` and `RecorderCocoa`

### LINKED ISSUE #11: Recorder onChange callback
Looking for an api to be notified when a recorder view successfully changes/removes its shortcut.  Something like:

```swift
KeyboardShortcuts.RecorderCocoa(for: .doSomething, onChange: { (shortcut: KeyboardShortcuts.Shortcut) in
 // do something with new shortcut key
})
```

I have my own preferences storage solution set up in an existing app and I want to update it when the recorder changes.  I suppose I'd like to disable the build-in UserDefaults storage as well, but not that's not a big deal.

gpoitch:
Adds ability to register callback when a Recorder view's value is changed/removed.

```swift
KeyboardShortcuts.Recorder(for: .doSomething, onChange: { (shortcut: KeyboardShortcuts.Shortcut?) in
  print(shortcut)
})
```
Closes #11

### DIFF
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -226,7 +226,6 @@ extension KeyboardShortcuts {
 			return shouldBecomeFirstResponder
 		}
 
-		/// :nodoc:
 		private func saveShortcut(_ shortcut: Shortcut?) {
 			if let shortcut = shortcut {
 				userDefaultsSet(name: shortcutName, shortcut: shortcut)
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -29,6 +29,7 @@ extension KeyboardShortcuts {
 		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
 		private let shortcutName: Name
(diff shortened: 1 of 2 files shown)


Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -49,8 +50,9 @@ extension KeyboardShortcuts {
 			}
 		}
 
-		public required init(for name: Name) {
+		public required init(for name: Name, onChange: ((_ shortcut: Shortcut?) -> Void)? = nil) {
```
sindresorhus, line 53: Can you add a documentation comment for `onChange`? Also in the other initializer.

Also mention the use-case for this (your use-case). Users don’t always realize why they would need a certain feature, and presenting some use-cases makes it clearer.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -205,13 +207,26 @@ extension KeyboardShortcuts {
...
+		private func saveShortcut(_ shortcut: Shortcut?) {
+			if let shortcut = shortcut {
+				userDefaultsSet(name: shortcutName, shortcut: shortcut)
+			} else {
+				userDefaultsRemove(name: shortcutName)
+			}
+
+			if let onChange = self.onChange {
+				onChange(shortcut)
+			}
```
sindresorhus, line 229: No need to if-let. Just `onChange?(...)`.

PR #13, opened 2020-06-16
Add ability to set default shortcut

### LINKED ISSUE #7: Shortcut default combination
Is it possible to add a default keys combination for the shortcut?

gpoitch:
Allows you to set a default key combination when defining a shortcut name.
```swift
extension KeyboardShortcuts.Name {
	static let toggleUnicornMode = Name("toggleUnicornMode", default: Shortcut(.a, modifiers: [.command]))
}
```
~Caveat: If you remove the default shortcut in the UI and restart the app, the default will always take over.  Would need to store a null value in user defaults or something.~

Closes #7

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -211,9 +211,13 @@ public enum KeyboardShortcuts {
 			return
 		}
 
-		UserDefaults.standard.removeObject(forKey: userDefaultsKey(for: name))
+		UserDefaults.standard.set(false, forKey: userDefaultsKey(for: name))
 		unregister(shortcut)
 		userDefaultsDidChange(name: name)
+	}
+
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Caveat: If you remove the default shortcut in the UI and restart the app, the default will always take over.

We cannot have that behavior. If the user clears the shortcut, it means they don't want it.

gpoitch:
> We cannot have that behavior. If the user clears the shortcut, it means they don't want it.

Yes, was hoping for a suggestion.

https://github.com/sindresorhus/KeyboardShortcuts/pull/13/commits/2fba62a97abcdfa0dde5f8fe05d1f1b60ac4a3b5 stores `false` instead of removing the key and that works.

sindresorhus:
Thanks for working on these improvements. Released: https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/v0.3.0

PR #24, opened 2020-08-08
Add `.setShortcut()`, `.getShortcut()`, and `.reset()` methods

### LINKED ISSUE #16: Manually set a keyboard shortcut for a name
This can be useful for migrating over to this package from some other solution.

Inside some one-time migration method:

```swift
let shortcut = KeyboardShortcuts.Shortcut(carbonKeycode: someHotkey, carbonModifiers: someModifiers)

KeyboardShortcuts.Name.foo.shortcut = shortcut
```

Alternatively:

```swift
let shortcut = KeyboardShortcuts.Shortcut(carbonKeycode: someHotkey, carbonModifiers: someModifiers)

KeyboardShortcuts.Name.set(shortcut, for: .foo)
```

That way, the user doesn't need to know the implementation details on how it's stored. Like here: https://github.com/p0deje/Maccy/commit/1001035d611cb411df702a941f5eb40626fc5ea7#diff-2589021fe69dedfc5619fc6a19136f23R117-R126

@p0deje Any thoughts?

### LINKED ISSUE #19: Add ability to reset a shortcut
This is a very useful package and the SwiftUI bindings make it a breeze to use. Please also add the ability to reset the shortcuts. I've tried resetting in the UserDefaults.standard directly but it didn't seem to have any effect.

I see that the code already has this ability, just that it needs to be exposed as a public API, say `KeyboardShortcuts.reset(name: Name)`

sindresorhus:
Fixes #19
Fixes #16

I'm looking for feedback on the docs and API. And if you have time, it would be great if you could test this in a real app.


sindresorhus:
One open question.

For:

```swift
extension KeyboardShortcuts.Name {
	static let foo = Self("foo", default: .init(.a, modifiers: .command))
}
```

Would you expect `KeyboardShortcuts.reset(.foo)` to reset the shortcut to `nil` or to `Command+A`?

sindresorhus:
// @pavanpodila @L1cardo @gpoitch @p0deje

pavanpodila:
Good point. I think you should reset to the default value, whatever it may be at the time of creating the shortcut. So in your case it should be `Command+A`

pavanpodila:
@sindresorhus can we get a new release :-)

sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/v0.5.0

pavanpodila:
Awesome! Thanks!

PR #25, opened 2020-08-10
Pause keyboard shortcut events while a recorder input is active

### LINKED ISSUE #20: Disable all shortcuts and then re-enable them
There should be a way to disable all shortcuts and then re-enable them. This is useful on a Preferences screen where you are registering a bunch of shortcuts. While you are entering new shortcuts, you should not be able to invoke the global shortcuts, especially when the control has focus. 

Currently there is no way to disable global hotkey monitoring.

sindresorhus:
Fixes #20

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -16,6 +16,9 @@ public enum KeyboardShortcuts {
 	private static var userDefaultsKeyDownHandlers = [Name: [KeyAction]]()
 	private static var userDefaultsKeyUpHandlers = [Name: [KeyAction]]()
 
+	/// When `true`, event handlers will not be called for registered keyboard shortcuts.
+	static var isPaused = false
+
 	private static func register(_ shortcut: Shortcut) {
 		guard !registeredShortcuts.contains(shortcut) else {
 			return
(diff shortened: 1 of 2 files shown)


sindresorhus:
@pavanpodila Can you try this PR in your app and see if it fixes the problem for you? You shouldn't have to do any code changes.

pavanpodila:
I'm using via SPM. Is there a temp version you can publish ?

sindresorhus:
SPM supports targeting branches and commits.

pavanpodila:
Ok, after switching to the recorder-pause branch, I am getting this strange error on using `Key.one` ?
```
(17, 67) 'one' is inaccessible due to 'internal' protection level
```

pavanpodila:
I think you may need to mark the extension on `KeyboardShortcuts.Key` as **`public`** ?

sindresorhus:
That will be fixed by the other open PR. Try this branch https://github.com/sindresorhus/KeyboardShortcuts/tree/rec which contains both PRs.

pavanpodila:
The rec branch seems to work well. When active, it correctly disables existing shortcuts 👍

sindresorhus:
Thanks for testing 👍🏻

pavanpodila:
Thanks for fixing this! Is there going to be a new release ?

sindresorhus:
When the other open PR is merged. But as you already know, you can target a specific commit if you need it right away.

PR #26, opened 2020-08-26
`Recorder` should allow changing the `name` parameter

hank121314:
## Summary

I found that 
```
KeyboardShortcuts.Recorder(for: shortCut)
```
If we changed `shortCut` in swiftUI the `shortcutName` name in RecorderCocoa will not update.
And I think we should also remove escape function in userDefaults when disalbed

## Test Plan

I edit KeyboardShortcutsExample to a dynamic shortCut name to test this situation.

Thanks you so much for your code review!


sindresorhus:
> And I think we should also remove escape function in userDefaults when disalbed

Why?

sindresorhus:
It's `Shortcut`, not `ShortCut` (applies everywhere).

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -28,7 +28,7 @@ extension KeyboardShortcuts {
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
 		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
-		private let shortcutName: Name
+		private var shortcutName: Name
```
sindresorhus, line 31: Instead of a `updateShortCutName` method, you can just make this property public and use `didSet`.

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -6,42 +6,57 @@ extension KeyboardShortcuts.Name {
...
-					self.isPressed2 = true
+				KeyboardShortcuts.onKeyDown(for: self.shortCuts[self.selectedState].name) {
+					self.isPressed = true
 				}
-
-				KeyboardShortcuts.onKeyUp(for: .testShortcut2) {
-					self.isPressed2 = false
+				KeyboardShortcuts.onKeyUp(for: self.shortCuts[self.selectedState].name) {
+					self.isPressed = false
 				}
```
sindresorhus, line 60: Instead of replacing the existing shortcut inputs here, can you make it an addition and put them below the existing ones, with a `Divider()` between?

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -6,42 +6,57 @@ extension KeyboardShortcuts.Name {
...
 
 	var body: some View {
 		VStack {
-			HStack {
-				KeyboardShortcuts.Recorder(for: .testShortcut1)
-					.padding(.trailing, 10)
-				Text("Pressed? \(isPressed1 ? "👍" : "👎")")
-					.frame(width: 100, alignment: .leading)
+			Picker(selection: selectedShortCut, label: Text("Select shortcut :")) {
+				ForEach(0 ..< shortCuts.count) {
```
sindresorhus, line 41: Don't loop based on index. You can just loop based on the elements.

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -6,42 +6,57 @@ extension KeyboardShortcuts.Name {
...
+	}
 
 	var body: some View {
 		VStack {
-			HStack {
-				KeyboardShortcuts.Recorder(for: .testShortcut1)
-					.padding(.trailing, 10)
-				Text("Pressed? \(isPressed1 ? "👍" : "👎")")
-					.frame(width: 100, alignment: .leading)
+			Picker(selection: selectedShortCut, label: Text("Select shortcut :")) {
```
sindresorhus, line 40:
```suggestion
			Picker(selection: selectedShortCut, label: Text("Select shortcut:")) {
```

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -6,42 +6,57 @@ extension KeyboardShortcuts.Name {
...
+	let label: String
+}
+
 struct ContentView: View {
-	@State private var isPressed1 = false
-	@State private var isPressed2 = false
+	@State private var isPressed = false
+	@State private var selectedState = 0
+	var shortCuts = [
+		ShortCut(name: .testShortcut1, id: 0, label: "ShortCut1"),
```
sindresorhus, line 19: No need for `id`. Can just use the array index for that.

KeyboardShortcutsExample/ContentView.swift
sindresorhus: And I think some of the code additions here could be simplified.

hank121314:
> > And I think we should also remove escape function in userDefaults when disalbed
> 
> Why?

I had met some issue before if I didn't remove this function, but now it seems to work fine.I discard that change.

KeyboardShortcutsExample/ContentView.swift
hank121314: > It's `Shortcut`, not `ShortCut` (applies everywhere).

OK! Sorry for my bad English 😢

hank121314:
> It's `Shortcut`, not `ShortCut` (applies everywhere).

OK! Sorry for my bad English 😢

hank121314:
I have pushed a new commit.

I am sorry for taking your time and thanks you for your code review again!

sindresorhus:
> I am sorry for taking your time and thanks you for your code review again!

Don't worry at all. Just pointing things out.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -28,7 +28,15 @@ extension KeyboardShortcuts {
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
 		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
-		private let shortcutName: Name
+		public var shortcutName: Name {
+			didSet {
```
sindresorhus, line 32: `didSet` is not called when we assign in an initializer. So we need to move the logic here to a private method and also call that method in the init.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -28,7 +28,15 @@ extension KeyboardShortcuts {
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
 		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
-		private let shortcutName: Name
+		public var shortcutName: Name {
```
sindresorhus, line 31: I think this should have a short doc comment pointing out that it can be dynamically changed. I don't think that's obvious.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -28,7 +28,15 @@ extension KeyboardShortcuts {
...
+			didSet {
+				if let shortcut = userDefaultsGet(name: shortcutName) {
+					self.stringValue = "\(shortcut)"
+				} else {
+					self.stringValue = ""
+				}
```
sindresorhus, line 37:
```suggestion
				stringValue = userDefaultsGet(name: shortcutName).map { "\($0)" } ?? ""
```

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -4,9 +4,75 @@ import KeyboardShortcuts
...
+	static let testShortcut4 = Self("testShortcut4")
 }
 
-struct ContentView: View {
+struct Shortcut {
+	let name: KeyboardShortcuts.Name
+	let label: String
+}
+
+class PickerShortcutViewModel: ObservableObject {
```
sindresorhus, line 16:
```suggestion
final class PickerShortcutViewModel: ObservableObject {
```

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -4,9 +4,75 @@ import KeyboardShortcuts
...
+	init() {
+		setShortcutEvent(name: shortcuts[selectedState].name)
+	}
+
+	var shortcuts = [
+		Shortcut(name: .testShortcut3, label: "Shortcut3"),
+		Shortcut(name: .testShortcut4, label: "Shortcut4")
+	]
+
+	func setShortcutEvent (name: KeyboardShortcuts.Name) {
```
sindresorhus, line 42:
```suggestion
	func setShortcutEvent(name: KeyboardShortcuts.Name) {
```

sindresorhus (changes requested):

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -4,9 +4,79 @@ import KeyboardShortcuts
...
+			self.isPressed = false
+		}
+	}
+}
+
+struct PickerShortcut: View {
+	@ObservedObject private var viewModel = PickerShortcutViewModel()
+
+	var body: some View {
+		VStack {
```
sindresorhus, line 60: I think this should have a header that says "Dynamic recorder" to make it clear that the example shows off how to use dynamic recorders.

sindresorhus:
Also mention the dynamic example in https://github.com/sindresorhus/KeyboardShortcuts#dynamic-keyboard-shortcuts

sindresorhus:
Can you fix the merge conflict?

sindresorhus:
By the way, I just left Taipei 3 weeks ago. I was in Taiwan for 5 months. I really liked it there.

hank121314:
> Can you fix the merge conflict?

Yes! I have pushed a new commit with merged from master.

hank121314:
> By the way, I just left Taipei 3 weeks ago. I was in Taiwan for 5 months. I really liked it there.

Wow! That's awesome! I look forward to welcoming you again! Maybe I can be your tour guide if your come to Taiwan next time 😄 !

hank121314:
Hi @sindresorhus !
Sorry for pinging you. Could you review this pr again?
Thank you very much!

sindresorhus:
Notice how the label is cut off when you select "Shortcut 4". That points to something not being updated correctly:

Screen Shot 2020-09-24 at 12 54 50

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -28,7 +28,12 @@ extension KeyboardShortcuts {
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
 		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
-		private let shortcutName: Name
+		/// shortcutName should be updated by NSViewRepresentable.updateView from Recorder and same as stringValue
```
sindresorhus, line 31: The comment should be a doc comment about what it does for the user, not mention internal implementation details.

readme.md
```diff
@@ -130,6 +130,24 @@ See [`NSMenuItem#setShortcut`](https://sindresorhus.com/KeyboardShortcuts/Extens
...
+struct PreferencesView: View {
+	@Binding var name: KeyboardShortcuts.Name
+	var body: some View {
+		HStack {
+			Text("Toggle Unicorn Mode:")
+			KeyboardShortcuts.Recorder(for: name)
+		}
+	}
+}
+```
```
sindresorhus, line 147: I think it would be better to move this to a doc comment in the actual code and link to that instead in https://sindresorhus.com/KeyboardShortcuts/Enums/KeyboardShortcuts/Recorder.html

KeyboardShortcutsExample/ContentView.swift
```diff
@@ -4,9 +4,102 @@ import KeyboardShortcuts
...
+		HStack {
+			KeyboardShortcuts.Recorder(for: name)
+				.padding(.trailing, 10)
+			Text("Pressed? \(isPressed ? "👍" : "👎")")
+				.frame(width: 100, alignment: .leading)
+		}
+	}
+}
+
+final class DynamicShortcutViewModel: ObservableObject {
```
sindresorhus, line 29: I think using a ViewModel for this is premature. This logic could be massively simplified and moved to the `DynamicShortcut` view.

sindresorhus (changes requested):

hank121314:
I have pushed a new commit!

But I have noticed some bugs.
If focus on Recorder and use picker to switch from empty stringValue to some stringValue.
The text might be cut off by placeholder.
Like the video below:
https://streamable.com/87vb9z

So I need to call blur() here: https://github.com/sindresorhus/KeyboardShortcuts/blob/6f1bb559533f254b1c1581ec9e96e056cc78d658/Sources/KeyboardShortcuts/RecorderCocoa.swift#L32-L39

Thank you for your kind assistance!

sindresorhus:
When I change a shortcut in the dropdown, it takes a second before the value changes. I think there's a problem with your state logic.

sindresorhus:
The "dynamic recorder" heading is not aligned properly:

Screen Shot 2020-10-05 at 22 15 44

hank121314:
> When I click the dropdown, select a different shortcut and then close the dropdown, it takes a second before the value changes. I think there's a problem with your state logic.

Unfortunately, I cannot reproduce this issue with my computer in Mac OS X 10.15.7.
But I still refactor the state logic.
Could you give me your system version or how to reproduce it?
Thanks!

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -48,7 +48,16 @@ extension KeyboardShortcuts {
...
+		public func updateNSView(_ nsView: NSViewType, context: Context) {
+			guard nsView.shortcutName != name else {
+				return
+			}
+			nsView.shortcutName = name
+			DispatchQueue.main.async {
+				// Wait for next cycle, so field be already in window
+				nsView.blur()
+			}
+		}
```
sindresorhus, line 60: Both these changes should be in the Cocoa view as they would be useful there too.

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -48,7 +48,16 @@ extension KeyboardShortcuts {
...
 		/// :nodoc:
-		public func updateNSView(_ nsView: NSViewType, context: Context) {}
+		public func updateNSView(_ nsView: NSViewType, context: Context) {
+			guard nsView.shortcutName != name else {
+				return
+			}
+			nsView.shortcutName = name
+			DispatchQueue.main.async {
+				// Wait for next cycle, so field be already in window
+				nsView.blur()
```
sindresorhus, line 58: The comment should instead describe why "blur" is needed.

sindresorhus (changes requested):

sindresorhus:
Same macOS version. Xcode 12.

hank121314:
> Same macOS version. Xcode 12.


I still can't reproduce the issue 😢 .
I wonder if this might cause by accessing UserDefaults too often ?
Because [getShortcut](https://github.com/sindresorhus/KeyboardShortcuts/blob/59f2f47241d1ec98f1df13d20bd4206347823b65/Sources/KeyboardShortcuts/KeyboardShortcuts.swift#L151-L160) will get call very often.
Would it help if we add a cache to store shortcut in `getShortcut`?
Like this:
```
private static var cacheShortcuts = [Name: Shortcut]()

public static func getShortcut(for name: Name) -> Shortcut? {
		guard cacheShortcuts[name] == nil else {
			return cacheShortcuts[name]
		}
		guard
			let data = UserDefaults.standard.string(forKey: userDefaultsKey(for: name))?.data(using: .utf8),
			let decoded = try? JSONDecoder().decode(Shortcut.self, from: data)
		else {
			return nil
		}

		cacheShortcuts[name] = decoded
		return decoded
}
```
Thank you for your suggestions!

sindresorhus:
No, it doesn't have anything to do with UserDefaults performance. UserDefaults is fast as it's in-memory and even if it was slow, it would not be "1 second" slow. This is definitely some state problem.

sindresorhus:
I cannot seem to reproduce this anymore, so it might have been fixed by one of the commits.

PR #27, opened 2020-08-27
Add localization support

### LINKED ISSUE #3: Localizations
Would be nice to ship with localizations:
- The recorder button has a couple of words.
- The alert when the keyboard shortcut is taken by the system or by the app's main menu.

However, since we support Swift Package Manager, we cannot do this until Swift 5.3, which [supports resources](https://github.com/apple/swift-evolution/blob/master/proposals/0271-package-manager-resources.md).

L1cardo:
Added localization support, and they look great. Fixes #3 

image
image
image


Sources/KeyboardShortcuts/Localization/String++.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Created by Licardo on 2020/8/28.
+//
+
+import Foundation
+
+extension String {
+    var localized: String {
+        NSLocalizedString(self, bundle: .module, comment: self)
+    }
+}
```
sindresorhus, line 14: I prefer to have this in the `util.swift` file.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -186,7 +186,7 @@ extension KeyboardShortcuts {
 
 					NSAlert.showModal(
 						for: self.window,
-						message: "This keyboard shortcut cannot be used as it's already used by the “\(menuItem.title)” menu item."
+                        message: "This keyboard shortcut cannot be used as it's already used by the".localized + " “\(menuItem.title)” " + "menu item.".localized
```
sindresorhus, line 189: This should be a single string. I think the localization stuff supports interpolation.

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
```diff
@@ -0,0 +1,6 @@
+"Record Shortcut" = "记录快捷键";
+"Press Shortcut" = "键入快捷键";
+"This keyboard shortcut cannot be used as it's already used by the" = "当前快捷键无法使用，因为它已用作菜单项";
+"menu item." = "的快捷键。";
+"This keyboard shortcut cannot be used as it's already a system-wide keyboard shortcut." = "当前快捷键无法使用，因为它已用作系统快捷键。";
+"Most system-wide keyboard shortcuts can be changed in “System Preferences › Keyboard › Shortcuts“." = "可以在 “系统偏好设置 › 键盘 › 快捷键” 中更改大多数系统快捷键。";
```
sindresorhus, line 6: Do you know anyone that knows Chinese that can review your translation?

Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
```diff
@@ -0,0 +1,6 @@
+"Record Shortcut" = "Record Shortcut";
+"Press Shortcut" = "Press Shortcut";
+"This keyboard shortcut cannot be used as it's already used by the" = "This keyboard shortcut cannot be used as it's already used by the";
+"menu item." = "menu item.";
+"This keyboard shortcut cannot be used as it's already a system-wide keyboard shortcut." = "This keyboard shortcut cannot be used as it's already a system-wide keyboard shortcut.";
+"Most system-wide keyboard shortcuts can be changed in “System Preferences › Keyboard › Shortcuts“." = "Most system-wide keyboard shortcuts can be changed in “System Preferences › Keyboard › Shortcuts“.";
```
sindresorhus, line 6: Do we really need to have a file that just duplicates the source?

sindresorhus:
Thanks for working on this. See the inline lint annotations.

sindresorhus:
Can you mention localization support in the readme?

Sources/KeyboardShortcuts/Localization/String++.swift
L1cardo: Done

Sources/KeyboardShortcuts/RecorderCocoa.swift
L1cardo: I have tried a single string, and it seems that Chinese localization does not support interpolation. I do not know why :(

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
L1cardo: Actually I am Chinese. :p

Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
L1cardo: This is my coding habit, which makes file structure more clear.

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
sindresorhus: Yes, I'm not doubting that. But it's always helpful to have someone else look over the translation to spot typos, etc. I would ask the same for any other language.

Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
sindresorhus: But they can just look at any other translation and look at the source strings there.

Sources/KeyboardShortcuts/RecorderCocoa.swift
sindresorhus: I know it's possible, so I would recommend Googling some more ;)

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
L1cardo: Yeah sure, I will let my friends have a look.

m01i0ng: @sindresorhus Chinese is my native language and I think this is the proper translation.

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -55,6 +55,11 @@ extension KeyboardShortcuts {
 @available(macOS 10.15, *)
 struct SwiftUI_Previews: PreviewProvider {
 	static var previews: some View {
-		KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+        Group {
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+                .environment(\.locale, .init(identifier: "en"))
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+                .environment(\.locale, .init(identifier: "zh-Hans"))
+        }
```
sindresorhus, line 63: Tab-indentation.

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
sindresorhus: Thanks for reviewing :)

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -63,7 +63,7 @@ extension KeyboardShortcuts {
 
 			super.init(frame: .zero)
 			self.delegate = self
-			self.placeholderString = "Record Shortcut"
+            self.placeholderString = "Record Shortcut".localized
```
sindresorhus, line 66: Tab-indentation

readme.md
```diff
@@ -116,6 +116,16 @@ final class PreferencesViewController: NSViewController {
...
 
+## Localization
+
+This package supports localization but still in early development, so PRs are welcome!
+
+1. Fork the repo.
+2. Create a directory that has a name that uses an ISO 639 language code and optional designators, followed by the `.lproj` suffix, and doesn’t contain subdirectories, you can find more [here](https://developer.apple.com/documentation/swift_packages/localizing_package_resources).
+3. Create a file named `Localizable.strings` under the new language directory, and copy all text from `KeyboardShortcuts/Localization/en.lproj/Localizable.strings` to the new file that you created.
+4. Localize!
+5. Submit a PR.
```
sindresorhus, line 127: Can you add something like this as step 6: Try to find someone that speaks your language to review the translation.

readme.md
```diff
@@ -116,6 +116,16 @@ final class PreferencesViewController: NSViewController {
...
 ```
 
+## Localization
+
+This package supports localization but still in early development, so PRs are welcome!
+
+1. Fork the repo.
+2. Create a directory that has a name that uses an ISO 639 language code and optional designators, followed by the `.lproj` suffix, and doesn’t contain subdirectories, you can find more [here](https://developer.apple.com/documentation/swift_packages/localizing_package_resources).
+3. Create a file named `Localizable.strings` under the new language directory, and copy all text from `KeyboardShortcuts/Localization/en.lproj/Localizable.strings` to the new file that you created.
+4. Localize!
```
sindresorhus, line 126:
```suggestion
4. Localize! And review your localization multiple times. Check for typos.
```

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
L1cardo: @sindresorhus Used a single string, please have a look if it is OK.

Sources/KeyboardShortcuts/RecorderCocoa.swift
L1cardo: image
This looks great on my Xcode, that is strange.

sindresorhus:
You also need to fix the SwiftLint annotations, ensure you use tab-indentation everywhere, remove the `.DS_Store` files. Also look over the diff after every change to ensure everything is correct.

Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
L1cardo: Deleted

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -186,7 +186,7 @@ extension KeyboardShortcuts {
 
 					NSAlert.showModal(
 						for: self.window,
-						message: "This keyboard shortcut cannot be used as it's already used by the “\(menuItem.title)” menu item."
+                        message: String(format: "This keyboard shortcut cannot be used as it's already used by the “%@” menu item.".localized, menuItem.title)
```
L1cardo, line 189: Is not this aligned vertically?
image

Sources/KeyboardShortcuts/RecorderCocoa.swift
sindresorhus: Install https://github.com/sindresorhus/refined-github and look at the diff and you'll see the difference in indentation. You could also turn on showing invisible in Xcode preferences.

Sources/KeyboardShortcuts/Localization/zh-Hans.lproj/Localizable.strings
Cub0ne: @sindresorhus I think this translation is just right,well done

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -186,7 +186,7 @@ extension KeyboardShortcuts {
 
 					NSAlert.showModal(
 						for: self.window,
-						message: "This keyboard shortcut cannot be used as it's already used by the “\(menuItem.title)” menu item."
+						message: String(format: "This keyboard shortcut cannot be used as it's already used by the “%@” menu item.".localized, menuItem.title)
```
sindresorhus, line 189: It looks like the most prop solution is using `localizedStringWithFormat`.

https://stackoverflow.com/a/47389291/64949
https://www.raywenderlich.com/250-internationalizing-your-ios-app-getting-started#toc-anchor-003

Sources/KeyboardShortcuts/util.swift
```diff
@@ -2,6 +2,14 @@ import Cocoa
 import Carbon.HIToolbox
 
 
+extension String {
+	/// localization
```
sindresorhus, line 6: Needs a better doc comment.

readme.md
```diff
@@ -116,6 +116,17 @@ final class PreferencesViewController: NSViewController {
 }
 ```
 
+## Localization
+
+This package supports localization but still in early development, so PRs are welcome!
+
```
sindresorhus, line 122: I think we should link to the `Localization` directory to show what localizations we currently support.

readme.md
```diff
@@ -116,6 +116,17 @@ final class PreferencesViewController: NSViewController {
...
+1. Fork the repo.
+2. Create a directory that has a name that uses an ISO 639 language code and optional designators, followed by the `.lproj` suffix, and doesn’t contain subdirectories, you can find more [here](https://developer.apple.com/documentation/swift_packages/localizing_package_resources).
+3. Create a file named `Localizable.strings` under the new language directory, and copy all text from `KeyboardShortcuts/Localization/en.lproj/Localizable.strings` to the new file that you created.
```
sindresorhus, line 125: `en.lproj` doesn't exist anymore.

Sources/KeyboardShortcuts/RecorderCocoa.swift
L1cardo: Done

L1cardo:
@sindresorhus All problems have been fixed.
The commits on this PR might be a little too many, I can re-PR with less commits if you want.

.gitignore
```diff
@@ -2,3 +2,4 @@
 /Packages
 xcuserdata
 project.xcworkspace
+.DS_Store
```
sindresorhus, line 5: This should be in your own global gitignore, not here.

Sources/KeyboardShortcuts/util.swift
```diff
@@ -3,7 +3,7 @@ import Carbon.HIToolbox
 
 
 extension String {
-	/// localization
+	/// Localize a string
```
sindresorhus, line 6: It's not "a" string, it's "the" string or "this" string.

A better description would be:

> Makes the string localizable.

L1cardo:
@sindresorhus Is there any problem left to be fixed?

sindresorhus:
I will take a final look in a couple of days. There's really no rush since this cannot be released until Xcode 12 GM is out anyway, which is at least 2 weeks away.

sindresorhus:
I tried building the example app on Xcode 12 beta 6 and Big Sur beta 6 and I'm getting an error:

```
KeyboardShortcuts/Sources/KeyboardShortcuts/util.swift:8:36: Type 'Bundle' has no member 'module'
```

L1cardo:
Here are the resources I found about this and I exactly followed.

https://developer.apple.com/wwdc20/10169
image
https://developer.apple.com/documentation/swift_packages/localizing_package_resources
image

L1cardo:
Another strange thing, I used this package in another test project, and I got no error.In example app, I got the same error.

L1cardo:
And I think we should not delete `en.lproj`.
image

sindresorhus:
Are you sure you're following the correct directory namings and structure?

sindresorhus:
> And I think we should not delete en.lproj.

I looked at some other projects and they use identifiers instead of just copying the sentence. So I think we should identifiers like `"FOO_BAR"` (find some suitable names) in the source and then also have the `en` localization.

L1cardo:
> > And I think we should not delete en.lproj.
> 
> 
> 
> I looked at some other projects and they use identifiers instead of just copying the sentence. So I think we should identifiers like `"FOO_BAR"` (find some suitable names) in the source and then also have the `en` localization.

I think direct copying will be more intuitive. Besides these are just a few sentences, not to much.

And I need to prepare for my postgraduate exam, this may hold for a while.

sindresorhus:
> I think direct copying will be more intuitive. Besides these are just a few sentences, not to much.

The problem is that if we do a tiny change in the English version, all translations has to be updated as their ID is the sentence itself. That's the benefit of constants and why I would prefer that. Also to reduce duplication since we need an English localization file anyway.

L1cardo:
> > I think direct copying will be more intuitive. Besides these are just a few sentences, not to much.
> 
> The problem is that if we do a tiny change in the English version, all translations has to be updated as their ID is the sentence itself. That's the benefit of constants and why I would prefer that. Also to reduce duplication since we need an English localization file anyway. It's not about the number of localized sentences, but I would rather find a good way to do it which I can use for other projects too. So let's go with identifiers.

This makes sense, identifiers then.

L1cardo:
@sindresorhus 
1. used identifier for localization instead of the sentence itself
2. added `en` localization

sindresorhus:
I'm still getting the "Type 'Bundle' has no member 'module'" error when building the example app, even on Xcode 12 GM. That needs to be fixed.

L1cardo:
> I'm still getting the "Type 'Bundle' has no member 'module'" error when building the example app, even on Xcode 12 GM. That needs to be fixed.

Start a new project and use online Swift Package, I see no error.
When you build the example project, it uses local Swift Package, and there will be error.
Don't know why.

sindresorhus:
My theory is that it's caused by the example app not consuming `KeyboardShortcuts` through Swift Package Manager (since SPM is the one making the `bundle.module` thing). We could exact the example app into a separate Xcode project, but I have a feeling it currently wouldn't work for Carthage/CocoaPods either then.

- https://developer.apple.com/forums/thread/650158
- https://developer.apple.com/forums/thread/652736

L1cardo:
Can you do the rest?
I really need to prepare for my exam.

sindresorhus:
I don't have time or interest to finish this, but there's no rush. I can leave it open until you have time.

sindresorhus:
This would be an ok workaround: https://github.com/AnyImageProject/AnyImageKit/blob/b3a8f2c91a516dde2cbb8a7e563e25532f3d54b5/Sources/AnyImageKit/Core/Extension/Core%2BBundle.swift

L1cardo:
OK I am back, done with my exam , let me see what is going on here.

L1cardo:
@sindresorhus I have tested this on Xcode 12.4, it seems all work fine.
As for Carthage/CocoaPods, I am not a fan of Carthage or CocoaPods, so I will not work on them.
Close or merge, do as you want, I really don't have time for this, thanks for understanding.

sindresorhus:
> As for Carthage/CocoaPods, I am not a fan of Carthage or CocoaPods

Me neither. I'm just going to drop support for them. People using those can still use SPM for this package.

sindresorhus:
Thanks for working on this :)

L1cardo:
> Thanks for working on this :)

Thank you too!  😃

Issue #1, opened 2020-05-07
Support menu bar apps that uses `NSMenu`


sindresorhus:
It currently doesn't work for menu bar that uses `NSMenu`. This is because `NSMenu` puts the event loop into a special tracking mode when the menu is open, which means the keyboard shortcut event will only arrive when the menu closes.

It should be possible to implement this by setting up a Carbon local event monitor instead when the menu is open. I have already made an swifty wrapper for it: https://gist.github.com/sindresorhus/2bb90276ad608a22ee5e8fb291b35b88 So we just need to use that instead when `NSMenu` is open.

The workaround for now is to disable the shortcut when the menu is open, register a local shortcut using the above code, and then when the menu closes, enable the global shortcut again.

**If you want to work on this, please read my comment in https://github.com/sindresorhus/KeyboardShortcuts/pull/28#issuecomment-706688451.**

PR #28, opened 2020-10-11
add CarbonLocalEventMonitor for NSMenu keyboard shortcut observation

honghaoz:
Not sure why you mentioned the manual way in #1 but didn't include it in the repo, would like to set up a shortcut observation when the menu is open.

This PR:
- adds the `CarbonLocalEventMonitor.swift` as mentioned in #1 
- adds a convenience method for observing shortcuts.

A example usage is like:
```swift
    let localEventMonitor = CarbonLocalEventMonitor.monitor(for: .toggleUnicornMode) {
      // handle the shortcut
    }

    menu.rx.menuOpenState
      .subscribe(onNext: { isOpen in
        if isOpen {
          KeyboardShortcuts.disable(.toggleUnicornMode)
          localEventMonitor.start()
        } else {
          KeyboardShortcuts.enable(.toggleUnicornMode)
          localEventMonitor.stop()
        }
      })
      .disposed(by: disposeBag)
```

### DIFF
--- a/Sources/KeyboardShortcuts/CarbonLocalEventMonitor.swift
+++ b/Sources/KeyboardShortcuts/CarbonLocalEventMonitor.swift
@@ -1,0 +1,194 @@
+import AppKit
+import Carbon
+
+// https://gist.github.com/sindresorhus/2bb90276ad608a22ee5e8fb291b35b88
+/**
+ Listen to local raw events using Carbon. The events are received before anything else gets access to them.
+ The given callback should return a boolean of whether the event was handled or not. If it's marked as handled, it's not propagated.
+ - Important: Don't forget to call `.start()`.
+ This class is especially useful for menu bar apps using NSMenu as events will still be available even in the event tracking run loop. `NSEvent.addGlobalMonitorForEvents` does not work in the event tracking run loop, so this is the only reliable way to receive events while NSMenu is open.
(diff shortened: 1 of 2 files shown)


sindresorhus:
It's easy enough to add the `CarbonLocalEventMonitor`. I could have done that myself too. The reason I didn't is that I would prefer a proper solution.

As I hinted at in the issue, the proper solution is automatic handling. I imagine we would provide a `KeyboardShortcuts.isMenuOpen` property, which the user could set to true/false when the menu is open/closed. And under the hood we would handle pausing the global keyboard shortcuts, activating the local monitor, etc.

We could also provide a `NSMenu` subclass which would handle calling `KeyboardShortcuts.isMenuOpen` automatically. We still need the property since not everyone can use a custom subclass.

Ideally, we would able to reuse https://github.com/sindresorhus/KeyboardShortcuts/blob/master/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift for local events too. I think it's just about what event specs we provide.

sindresorhus:
@honghaoz Are you still interested in working on this?

honghaoz:
Hi, @sindresorhus I'm not working on this one as I'm not an expert on Carbon APIs. I think we should close this PR.

PR #31, opened 2020-11-22
Added documentation of the `default` parameter.

daveverwer:
I found this through the pull request that added it, but it should be a little clearer so I thought it might fit in the readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -92,6 +92,17 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
 *You can also listen to key down with `.onKeyDown()`*
 
+You can provide a default key combination when defining the shortcut by passing a `default` parameter:
+
+```swift
+import KeyboardShortcuts
+
+extension KeyboardShortcuts.Name {
+	static let toggleUnicornMode = Self("toggleUnicornMode",
+	     default: Shortcut(.u, modifiers: [.command, .option]))
+}
+```
+
 **That's all! ✨**
 
 You can find a complete example by opening `KeyboardShortcuts.xcodeproj` and then running the `KeyboardShortcutsExample` target.


sindresorhus:
I intentionally didn't add it to the readme as I don't want people to reach for it. I fear that a lot of devs will use it incorrectly by predefining keyboard shortcuts for users. As a user, this is annoying as it might override some of my existing keyboard shortcuts. A lot of apps do this. Of course, there are some valid uses for it, like migration. Per:

> https://github.com/sindresorhus/KeyboardShortcuts/blob/af553c76a07ace761d159ebca5f4bd7df397d1ed/Sources/KeyboardShortcuts/Name.swift#L25

Just curious, what are you using it for?

daveverwer:
I understand (and agree with) your reasoning, but I'm not sure I agree that it shouldn't be documented because of that. I found the mechanism to do this, but didn't find your reasoning on why it wasn't mentioned, so I'm not sure the plan is working. 😅

Would it be better to document it, and include the reasoning on why to be cautious with it in the documentation? I'd be happy to adjust the PR to include that reasoning.

In terms of why I'm using it. It's for an app only for me (at the moment at least) so I feel comfortable with it. It's an app which really only makes sense from a keyboard shortcut, so it's an important part of it, but if I do ever release it, I'll make the keyboard shortcut setup part of an onboarding process.

sindresorhus:
> I found the mechanism to do this, but didn't find your reasoning on why it wasn't mentioned, so I'm not sure the plan is working.

It's documented both in the Xcode inspector when you select the type (like everything else) and also on the docs website: https://sindresorhus.com/KeyboardShortcuts/Enums/KeyboardShortcuts/Name.html#/s:17KeyboardShortcutsAAO4NameV_7defaultADSS_AB8ShortcutVSgtcfc

sindresorhus:
> In terms of why I'm using it. It's for an app only for me (at the moment at least) so I feel comfortable with it. It's an app which really only makes sense from a keyboard shortcut, so it's an important part of it, but if I do ever release it, I'll make the keyboard shortcut setup part of an onboarding process.

That's a valid use-case.

PR #36, opened 2021-01-07
Add method to remove all handlers

pvieito:
This is required if your keyboard shortcut names are dynamic (eg. they depend on user content) and you want to reset the handlers before re-creating them to avoid having multiple handlers for the same shortcut.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -45,7 +45,7 @@ public enum KeyboardShortcuts {
 
 		// TODO: Should remove user defaults too.
 	}
-	
+
 	/**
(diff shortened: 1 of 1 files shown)


Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -45,6 +45,16 @@ public enum KeyboardShortcuts {
 
 		// TODO: Should remove user defaults too.
 	}
+	
+	/**
+	Remove all handlers receiving keyboard shortcuts events.
```
sindresorhus, line 50: Can you also document your use-case for it as mention in the PR description? It's not always clear to users what they would use something for and one or more use-cases can help them realize.

pvieito: Added 👍🏻

PR #50, opened 2021-06-09
Support for controlSize(:) and frame(width:) modifier

Maschina:
My current app design requires a smaller variant of the text field. Thus, I have implemented a proper support for controlSize(:) that (among others) reflects the font size to the controlSize parameter. Furthermore, changing the text field's width was an important requirements for proper layouting.


Maschina:
FYI: Please refer to the Previews in Recorder.swift for testings.

PR #51, opened 2021-06-09
Support for controlSize(:) and frame(width:) modifier

Maschina:
My current app design requires a smaller variant of the text field. Thus, I have implemented a proper support for controlSize(:) that (among others) reflects the font size to the controlSize parameter. Furthermore, changing the text field's width was an important requirements for proper layouting.

FYI: Please refer to the Previews in Recorder.swift for testings.

### DIFF
--- a/Sources/KeyboardShortcuts/Recorder.swift
+++ b/Sources/KeyboardShortcuts/Recorder.swift
@@ -31,39 +31,70 @@ extension KeyboardShortcuts {
 
 		private let name: Name
 		private let onChange: ((_ shortcut: Shortcut?) -> Void)?
+        private let isHorizontalContentSizeConstraintActive: Bool
 
-		/**
-		- Parameter name: Strongly-typed keyboard shortcut name.
-		- Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
-		*/
+        /**
+         - Parameter name: Strongly-typed keyboard shortcut name.
+         - Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
+         - Parameter isHorizontalContentSizeConstraintActive: When `true`, makes sure to have width constraint set to avoid cutting placeholder string. When `false`, the frame(:) modifier will control the actual width of the text field.
+         */
 		public init(
 			for name: Name,
-			onChange: ((_ shortcut: Shortcut?) -> Void)? = nil
+			onChange: ((_ shortcut: Shortcut?) -> Void)? = nil,
+            constraint isHorizontalContentSizeConstraintActive: Bool = false
 		) {
 			self.name = name
 			self.onChange = onChange
+            self.isHorizontalContentSizeConstraintActive = isHorizontalContentSizeConstraintActive
 		}
 
 		/// :nodoc:
-		public func makeNSView(context: Context) -> NSViewType { .init(for: name, onChange: onChange) }
+		public func makeNSView(context: Context) -> NSViewType { .init(for: name, onChange: onChange, isHorizontalContentSizeConstraintActive: isHorizontalContentSizeConstraintActive) }
 
 		/// :nodoc:
 		public func updateNSView(_ nsView: NSViewType, context: Context) {
 			nsView.shortcutName = name
+            
+            // support for controlSize(:) modifier
+            switch context.environment.controlSize {
+                case .small:
+                    nsView.font = .systemFont(ofSize: 11)
+                    nsView.changeWidth(width: 110)
+                case .mini:
+                    nsView.font = .systemFont(ofSize: 9)
+                    nsView.changeWidth(width: 95)
+                default:
+                    break
+            }
 		}
 	}
 }
 
 @available(macOS 10.15, *)
 struct SwiftUI_Previews: PreviewProvider {
-	static var previews: some View {
+    static var previews: some View {
 		Group {
 			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
 				.environment(\.locale, .init(identifier: "en"))
-			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
-				.environment(\.locale, .init(identifier: "zh-Hans"))
-			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
-				.environment(\.locale, .init(identifier: "ru"))
+                .frame(width: 130)
+            
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+                .previewDisplayName("Test: .controlSize(.small)")
+                .environment(\.locale, .init(identifier: "en"))
+                .controlSize(.small)
+                .frame(width: 110)
+            
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+                .previewDisplayName("Test: .controlSize(.mini)")
+                .environment(\.locale, .init(identifier: "en"))
+                .controlSize(.mini)
+                .frame(width: 95)
+                
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"), constraint: true)
+                .previewDisplayName("Test: Horizontal constraint")
+                .environment(\.locale, .init(identifier: "en"))
+                .frame(width: 50)
 		}
+        .padding()
 	}
 }
--- a/Sources/KeyboardShortcuts/Recorder.swift
+++ b/Sources/KeyboardShortcuts/Recorder.swift
@@ -31,70 +31,66 @@ extension KeyboardShortcuts {
 
 		private let name: Name
 		private let onChange: ((_ shortcut: Shortcut?) -> Void)?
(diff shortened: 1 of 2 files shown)


sindresorhus:
I would prefer a solution that doesn't require introducing a new option.

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -31,39 +31,70 @@ extension KeyboardShortcuts {
 
 		private let name: Name
 		private let onChange: ((_ shortcut: Shortcut?) -> Void)?
+        private let isHorizontalContentSizeConstraintActive: Bool
```
sindresorhus, line 34: Even if we decide to introduce a new option, this option is too specific.

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -31,39 +31,70 @@ extension KeyboardShortcuts {
...
 
 		/// :nodoc:
 		public func updateNSView(_ nsView: NSViewType, context: Context) {
 			nsView.shortcutName = name
+            
+            // support for controlSize(:) modifier
+            switch context.environment.controlSize {
+                case .small:
+                    nsView.font = .systemFont(ofSize: 11)
+                    nsView.changeWidth(width: 110)
```
sindresorhus, line 62: Why manually apply this instead of just using `NSControl#controlSize`?

Maschina:
> I would prefer a solution that doesn't require introducing a new option.

You are referring to `isHorizontalContentSizeConstraintActive`?

Sources/KeyboardShortcuts/Recorder.swift
```diff
@@ -31,39 +31,70 @@ extension KeyboardShortcuts {
...
 			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
 				.environment(\.locale, .init(identifier: "en"))
-			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
-				.environment(\.locale, .init(identifier: "zh-Hans"))
-			KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
-				.environment(\.locale, .init(identifier: "ru"))
+                .frame(width: 130)
+            
+            KeyboardShortcuts.Recorder(for: .init("xcodePreview"))
+                .previewDisplayName("Test: .controlSize(.small)")
```
sindresorhus, line 82: Please use tab-indentation everywhere and make sure listing passes.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -67,10 +67,10 @@ extension KeyboardShortcuts {
...
 		}
 
-		/**
-		- Parameter name: Strongly-typed keyboard shortcut name.
-		- Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
-		*/
+        /**
+         - Parameter name: Strongly-typed keyboard shortcut name.
+         - Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
+         */
```
sindresorhus, line 73: Don't do unrelated changes.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/Recorder.swift
Maschina: Any counter proposal? The actual init parameter is called `constraint`.

Sources/KeyboardShortcuts/Recorder.swift
Maschina: Good point, will fix it

sindresorhus:
> You are referring to isHorizontalContentSizeConstraintActive?

Yes

Sources/KeyboardShortcuts/Recorder.swift
Maschina: We might go this approach: We completely get right of the built-in width constraint. The control is utilizing its width to the available space completely. 

Seeing how all other native SwiftUI controls are implemented, I would say that this behavior would be more SwiftUI'ful, because their widths is always controlled through either modifiers (e.g. `.frame()`), or ViewBuilder structs.
I haven't seen any native SwiftUI control that has a fixed width constraint.

What do you think?

sindresorhus: I partly agree. The control still needs a minimum width (and intrinsic content size) to make sure the label is not cut off.

Maschina: OK, I see your point. Nevertheless, this is a topic every UI designer has to handle with each and every other native controls such as password textfield etc. 

I see your point that the keyboard recorder control requires a minimum width.
Maybe we could control the minimum width in the `override public var intrinsicContentSize: CGSize { ... }` function by not allowing a width of the size below a certain value?

sindresorhus: > I agree with you that the keyboard recorder control requires a minimum width.
Maybe we could control the minimum width in the override public var intrinsicContentSize: CGSize { ... } function by not allowing a width of the size below a certain value?

👍

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -88,9 +92,9 @@ extension KeyboardShortcuts {
 			self.wantsLayer = true
 			self.translatesAutoresizingMaskIntoConstraints = false
 			setContentHuggingPriority(.defaultHigh, for: .vertical)
-			setContentHuggingPriority(.defaultHigh, for: .horizontal)
-			widthAnchor.constraint(greaterThanOrEqualToConstant: CGFloat(minimumWidth)).isActive = true
-
+			setContentHuggingPriority(.defaultLow, for: .horizontal)
+			self.isHorizontalContentSizeConstraintActive = true
```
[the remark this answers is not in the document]
sindresorhus: It also requires macOS 10.15, so it needs to be guarded.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -138,6 +142,15 @@ extension KeyboardShortcuts {
 				focus()
 			}
 		}
+		
+		public func changeWidth(for size: NSControl.ControlSize) {
```
sindresorhus, line 146: This should not be a public method. The control size should be handled like any other control. I guess, you could override `.controlSize` and handle the changes there.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -26,10 +26,14 @@ extension KeyboardShortcuts {
...
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
-		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
 		private let onChange: ((_ shortcut: Shortcut?) -> Void)?
 		private var observer: NSObjectProtocol?
+		private var intrinsicWidth: CGFloat = 130
+		
+		private let intrinsicWidthRegular: CGFloat = 130
+		private let intrinsicWidthSmall: CGFloat = 110
+		private let intrinsicWidthMini: CGFloat = 95
```
sindresorhus, line 36: We should not hard-code the minimum widths, since we support localization. They should be based on the intrinsic content size of the labels in the button, with some tiny padding on each side.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -26,10 +26,14 @@ extension KeyboardShortcuts {
 	```
 	*/
 	public final class RecorderCocoa: NSSearchField, NSSearchFieldDelegate {
-		private let minimumWidth: Double = 130
 		private var eventMonitor: LocalEventMonitor?
 		private let onChange: ((_ shortcut: Shortcut?) -> Void)?
 		private var observer: NSObjectProtocol?
+		private var intrinsicWidth: CGFloat = 130
```
sindresorhus, line 32: This is moot. Just use `intrinsicWidthRegular`.

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -68,9 +72,9 @@ extension KeyboardShortcuts {
 		}
 
 		/**
-		- Parameter name: Strongly-typed keyboard shortcut name.
-		- Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
-		*/
+		 - Parameter name: Strongly-typed keyboard shortcut name.
+		 - Parameter onChange: Callback which will be called when the keyboard shortcut is changed/removed by the user. This can be useful when you need more control. For example, when migrating from a different keyboard shortcut solution and you need to store the keyboard shortcut somewhere yourself instead of relying on the built-in storage. However, it's strongly recommended to just rely on the built-in storage when possible.
+		 */
```
sindresorhus, line 77: Always look over the diff after doing changes.

sindresorhus (changes requested):

sindresorhus:
You still have some lint warnings. Look at the diff.

sindresorhus:
You also need to update the example project and docs to use `.fixedSize()` to make it the intrinsic content size as that's what most users will want.

sindresorhus:
Closing for lack of activity.

PR #69, opened 2022-02-06
Add as SwiftUI `.keyboardShortcut()` helper

nighthawk:
This adds a `.keyboardShortcut(_ shortcut: KeyboardShortcuts.Name)` helper function to assign a shortcut defined through this `KeyboardShortcuts` to a SwiftUI button. This provides a similar functionality to the existing `NSMenuItem` extension, but when using SwiftUI's `Commands` to define your menu items. 

This can be used as follows:

```swift
@main
struct MyApp: App {
  
  var body: some Scene {
    WindowGroup {
      ...
    }
    .commands {
      CommandMenu("Unicorns") {
        Button("Toggle Unicorn Mode") { ... }
        .keyboardShortcut(.toggleUnicornMode)
      }
    }
  }

}
```

In my testing on macOS 12 this works well without having to disable the global keyboard shortcut while the menu is visible, but I'm not really sure if it's safe to do so.

### DIFF
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -300,8 +300,9 @@ extension KeyboardShortcuts.Shortcut: CustomStringConvertible {
 extension KeyboardShortcuts.Shortcut {
 	
 	@available(iOS 14.0, macOS 11.0, *)
-	var swiftUI: SwiftUI.KeyboardShortcut {
-		.init(.init(keyEquivalent.first!), modifiers: modifiers.swiftUI)
+	var swiftUI: SwiftUI.KeyboardShortcut? {
+		guard let key = keyEquivalent.first else { return nil }
+		return .init(.init(key), modifiers: modifiers.swiftUI)
 	}
 	
 }
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -298,55 +298,55 @@ extension KeyboardShortcuts.Shortcut: CustomStringConvertible {
 // MARK: - SwiftUI Helpers
 
 extension KeyboardShortcuts.Shortcut {
-  
-  @available(iOS 14.0, macOS 11.0, *)
-  var swiftUI: SwiftUI.KeyboardShortcut {
-    .init(.init(keyEquivalent.first!), modifiers: modifiers.swiftUI)
-  }
-
+	
+	@available(iOS 14.0, macOS 11.0, *)
+	var swiftUI: SwiftUI.KeyboardShortcut {
+		.init(.init(keyEquivalent.first!), modifiers: modifiers.swiftUI)
+	}
+	
 }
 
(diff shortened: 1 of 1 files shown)


Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -292,3 +294,59 @@ extension KeyboardShortcuts.Shortcut: CustomStringConvertible {
...
+@available(iOS 14.0, macOS 11.0, *)
+extension View {
+  
+  @ViewBuilder
+  /// Assigns the global keyboard shortcut to the modified control.
+  ///
+  /// Only assigns a keyboard shortcut, if one was defined (or it has a default shortcut).
+  ///
+  /// - Parameter shortcut: Strongly-typed name of the shortcut
+  public func keyboardShortcut(_ shortcut: KeyboardShortcuts.Name) -> some View {
```
sindresorhus, line 344: If the shortcut attached to `KeyboardShortcuts.Name` is changed by the user, it will not be changed here. You need to use a `ViewModifier` and listen to updates to the keyboard shortcut, like the `NSMenuItem` extension does.

Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -292,3 +294,59 @@ extension KeyboardShortcuts.Shortcut: CustomStringConvertible {
...
+    if let shortcut = (shortcut.shortcut ?? shortcut.defaultShortcut)?.swiftUI {
+      self.keyboardShortcut(shortcut)
+    } else {
+      self
+    }
+  }
```
sindresorhus, line 350: Use tab-indentation.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/Shortcut.swift
nighthawk: I expected so, too, but that's not necessary in my testing. As soon as I assign a new keyboard shortcut using the `Recorder`, the `.commands` is re-evaluated as this method will be called again.

Sources/KeyboardShortcuts/Shortcut.swift
nighthawk: Fixed.

Sources/KeyboardShortcuts/Shortcut.swift
sindresorhus: I think that only applies if they're in the same view. But that's not the case if they are in different windows (preferences vs main app window).

sindresorhus:
@nighthawk Still interested in working or this or should I close?

nighthawk:
Excuse the slow reply. I didn't yet have time to look into how to do it using a `ViewModifier` and haven't yet had any urgency for it as it works as is for how I use it even though I use it across different views, i.e., the app's settings view (to set it) and the app's main window's menu item (to display and trigger it).

sindresorhus:
Closing for lack of activity. https://github.com/sindresorhus/KeyboardShortcuts/issues/101

PR #71, opened 2022-02-22
Add Czech Localization

isametry:


sindresorhus:
Do you know anyone that would be able to help review your localization?

isametry:
Hmm, I don’t know… maybe @lemafor or @jezevec10 ?

Sources/KeyboardShortcuts/Localization/cs.lproj/Localizable.strings
```diff
@@ -0,0 +1,5 @@
+"record_shortcut" = "Přidat zkratku";
+"press_shortcut" = "Stiskněte klávesy";
+"keyboard_shortcut_used_by_menu_item" = "Tuto zkratku nelze použít, protože ji už využívá položka „%@“ v nabídce.";
```
junkycoder, line 3: "Tuto zkratku nelze použít, protože je již využívána položkou „%@“." imho sound better in this context.

Sources/KeyboardShortcuts/Localization/cs.lproj/Localizable.strings
```diff
@@ -0,0 +1,5 @@
...
+"keyboard_shortcut_used_by_system" = "Tuto zkratku nelze použít, protože už ji používá systém.";
```
junkycoder, line 4: 👌

Sources/KeyboardShortcuts/Localization/cs.lproj/Localizable.strings
```diff
@@ -0,0 +1,5 @@
...
+"keyboard_shortcuts_can_be_changed" = "Většinu systémových zkratek můžete změnit v „Předvolby systému › Klávesnice › Zkratky“.";
```
junkycoder, line 5: 👌

junkycoder (commented):
Already "looks good on dev", I just left few suggestions..

YassinEldeeb:
I think @capaj can help.

jezevec10 (approved):
Both the original and the suggested translations can be used, no mistakes found.

capaj (approved):
LGTM, but suggestions by @junkycoder are even better in both places.

isametry:
I also agree with the changes suggested by @junkycoder. In the first case, another possibility would be "Zadejte zkratku" or "Zadejte klávesy" – the latter is used by Apple in this context, and since this is a macOS package, staying consistent to the system would make the most sense I think.

press_shortcut

sindresorhus:
> the latter is used by Apple in this context, and since this is a macOS package, staying consistent to the system would make the most sense I think.

👍

sindresorhus:
@isametry You need to update your pull request with the changes ;)

sindresorhus:
Thank you everyone :)

PR #79, opened 2022-04-12
View Modifier Implementations

EmilioPelaez:
@sindresorhus I added an implementation using ViewModifiers to more closely match the syntax.

I'm not super sure about the API design, because I don't love the name of the functions. Maybe keeping only the implementation of `onShortcutEvent` but renamed to `onShortcut` would look cleaner.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -415,6 +415,7 @@ extension KeyboardShortcuts {
 				streamKeyDownHandlers[name, default: [:]][id] = {
 					continuation.yield(.keyDown)
 				}
+
 				streamKeyUpHandlers[name, default: [:]][id] = {
 					continuation.yield(.keyUp)
 				}
@@ -466,7 +467,7 @@ extension KeyboardShortcuts {
 	public static func events(_ type: EventType, for name: Name) -> AsyncFilterSequence<AsyncStream<EventType>> {
 		events(for: name).filter { $0 == type }
 	}
-	
+
 	@available(macOS 10.15, *)
 	@available(*, deprecated, renamed: "events(_:for:)")
 	public static func on(_ type: EventType, for name: Name) -> AsyncStream<Void> {
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -411,20 +411,24 @@ extension KeyboardShortcuts {
 		AsyncStream { continuation in
 			let id = UUID()
 
-			streamKeyDownHandlers[name, default: [:]][id] = {
-				continuation.yield(.keyDown)
-			}
-			streamKeyUpHandlers[name, default: [:]][id] = {
-				continuation.yield(.keyUp)
-			}
+			DispatchQueue.main.async {
+				streamKeyDownHandlers[name, default: [:]][id] = {
+					continuation.yield(.keyDown)
+				}
+				streamKeyUpHandlers[name, default: [:]][id] = {
+					continuation.yield(.keyUp)
+				}
 
-			registerShortcutIfNeeded(for: name)
+				registerShortcutIfNeeded(for: name)
+			}
 
 			continuation.onTermination = { _ in
-				streamKeyDownHandlers[name]?[id] = nil
-				streamKeyUpHandlers[name]?[id] = nil
+				DispatchQueue.main.async {
+					streamKeyDownHandlers[name]?[id] = nil
+					streamKeyUpHandlers[name]?[id] = nil
 
-				unregisterShortcutIfNeeded(for: name)
+					unregisterShortcutIfNeeded(for: name)
+				}
 			}
 		}
 	}
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -429,6 +429,35 @@ extension KeyboardShortcuts {
 		}
 	}
 
+	/**
+	Listen to keyboard shortcut events with the given name and type.
+
+	You can register multiple listeners.
+
+	You can safely call this even if the user has not yet set a keyboard shortcut. It will just be inactive until they do.
+
+	Ending the async sequence will stop the listener. For example, in the below example, the listener will stop when the view disappears.
+
+	```swift
+	import SwiftUI
+	import KeyboardShortcuts
+
+	struct ContentView: View {
+		@State private var isUnicornMode = false
+
+		var body: some View {
+			Text(isUnicornMode ? "🦄" : "🐴")
+				.task {
+					for await event in KeyboardShortcuts.events(for: .toggleUnicornMode) where event == .keyUp {
+						isUnicornMode.toggle()
+					}
+				}
+		}
+	}
+	```
+
+	- Note: This method is not affected by `.removeAllHandlers()`.
+	*/
 	@available(macOS 10.15, *)
 	public static func events(_ type: EventType, for name: Name) -> AsyncFilterSequence<AsyncStream<EventType>> {
 		events(for: name).filter { $0 == type }
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -430,7 +430,12 @@ extension KeyboardShortcuts {
 	}
 
 	@available(macOS 10.15, *)
-	@available(*, deprecated, message: "Use events(for:) instead")
+	public static func events(for name: Name, type: EventType) -> AsyncFilterSequence<AsyncStream<EventType>> {
+		events(for: name).filter { $0 == type }
+	}
+
+	@available(macOS 10.15, *)
+	@available(*, deprecated, renamed: "events(for:type:)")
 	public static func on(_ type: EventType, for name: Name) -> AsyncStream<Void> {
 		AsyncStream { continuation in
 			let id = UUID()
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -430,12 +430,12 @@ extension KeyboardShortcuts {
 	}
 
 	@available(macOS 10.15, *)
-	public static func events(for name: Name, type: EventType) -> AsyncFilterSequence<AsyncStream<EventType>> {
+	public static func events(_ type: EventType, for name: Name) -> AsyncFilterSequence<AsyncStream<EventType>> {
 		events(for: name).filter { $0 == type }
 	}
-
+	
 	@available(macOS 10.15, *)
-	@available(*, deprecated, renamed: "events(for:type:)")
+	@available(*, deprecated, renamed: "events(_:for:)")
 	public static func on(_ type: EventType, for name: Name) -> AsyncStream<Void> {
 		AsyncStream { continuation in
 			let id = UUID()
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -396,7 +396,7 @@ extension KeyboardShortcuts {
 		var body: some View {
 			Text(isUnicornMode ? "🦄" : "🐴")
 				.task {
-					for await _ in KeyboardShortcuts.on(.keyUp, for: .toggleUnicornMode) {
+					for await event in KeyboardShortcuts.events(for: .toggleUnicornMode) where event == .keyUp {
 						isUnicornMode.toggle()
 					}
 				}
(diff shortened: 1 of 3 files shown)


sindresorhus:
I think it should be named `onKeyboardShortcut`.

sindresorhus:
Maybe two methods. One where you specify the event type and one without that receives the type as parameter?

sindresorhus:
You don't need view modifiers for this. View extensions are enough.

EmilioPelaez:
@sindresorhus changes made

Sources/KeyboardShortcuts/ViewModifiers.swift
```diff
@@ -0,0 +1,27 @@
...
+				}
+			}
+			.task {
+				for await _ in KeyboardShortcuts.on(.keyUp, for: shortcut) {
+					perform(.keyUp)
+				}
+			}
+	}
+
+	public func onKeyboardShortcut(_ shortcut: KeyboardShortcuts.Name, event: KeyboardShortcuts.EventType, perform: @escaping () -> Void) -> some View {
```
sindresorhus, line 19:
```suggestion
	public func onKeyboardShortcut(_ shortcut: KeyboardShortcuts.Name, type: KeyboardShortcuts.EventType, perform: @escaping () -> Void) -> some View {
```

Sources/KeyboardShortcuts/ViewModifiers.swift
```diff
@@ -0,0 +1,27 @@
...
+		self
+			.task {
```
sindresorhus, line 21:
```suggestion
		task {
```

Sources/KeyboardShortcuts/ViewModifiers.swift
```diff
@@ -0,0 +1,27 @@
+import SwiftUI
+
+@available(macOS 12, *)
+extension View {
+	public func onKeyboardShortcut(_ shortcut: KeyboardShortcuts.Name, perform: @escaping (KeyboardShortcuts.EventType) -> Void) -> some View {
+		self
+			.task {
+				for await _ in KeyboardShortcuts.on(.keyDown, for: shortcut) {
```
sindresorhus, line 8: I wonder if it makes sense to make it possible to not specify an event type for `.on(` too. That way we wouldn't need two `.task`'s here. The problem is that `.on(for: shortcut)` is ambiguous when you read it... "on" for what.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/ViewModifiers.swift
EmilioPelaez: That's something I wanted to raise in another issue or PR, a single stream that returns the event type could make sense. I had already written it actually:

```
	@available(macOS 10.15, *)
	public static func events(for name: Name) -> AsyncStream<KeyboardShortcuts.EventType> {
		AsyncStream { continuation in
			let id = UUID()

			streamKeyDownHandlers[name, default: [:]][id] = {
				continuation.yield(.keyDown)
			}

			streamKeyUpHandlers[name, default: [:]][id] = {
				continuation.yield(.keyUp)
			}

			registerShortcutIfNeeded(for: name)

			continuation.onTermination = { _ in
				streamKeyDownHandlers[name]?[id] = nil
				streamKeyUpHandlers[name]?[id] = nil

				unregisterShortcutIfNeeded(for: name)
			}
		}
	}
```

sindresorhus: Having two similar methods with completely different names kind bothers me though. "on" vs "events".

Can you think of any way we could have it consistent?

EmilioPelaez: I looked at other implementations that return an `AsyncStream` to get an idea of the grammar they use:
```
for try await line in url.lines
for try await data in publisher.values
```
The properties are usually named after the value that the stream collects, rather than when the collection happens (like `on(shortcut)` is).

My recommendation would be to use the `events` name, since it more closely follows the existing pattern with APIs using `AsyncStream`.

Thoughts?

sindresorhus: > (Maybe deprecate the existing one and provide two implementations, one that returns all the events and one that receives an EventType parameter and only sends those kinds of events)

Yeah. That's the best solution.

EmilioPelaez: @sindresorhus I made the changes.

I also ran into a crash that I think is unrelated (it happens on the new `events` function and on the existing `on` function). I don't know if it's something you've seen before:
```
-[NSTaggedPointerString count]: unrecognized selector sent to instance 0x8000000000000000
```

<details>
<summary>Full Stack Trace</summary>

```
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001b110ab08 __exceptionPreprocess + 240
	1   libobjc.A.dylib                     0x00000001b0e55e14 objc_exception_throw + 60
	2   CoreFoundation                      0x00000001b119ea68 -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x00000001b1069e14 ___forwarding___ + 1764
	4   CoreFoundation                      0x00000001b1069670 _CF_forwarding_prep_0 + 96
	5   libswiftCore.dylib                  0x00000001be1b35b0 $sSD8_VariantV8setValue_6forKeyyq_n_xtF + 120
	6   libswiftCore.dylib                  0x00000001be4386c8 $sSDyq_SgxcisTf4ngn_n + 428
	7   libswiftCore.dylib                  0x00000001be169f8c $sSDyq_Sgxcis + 32
	8   KeyboardShortcutsExample            0x0000000102184b00 $s17KeyboardShortcuts06CarbonaB0O8register_9onKeyDown0eF2UpyA2AO8ShortcutV_yAIcyAIctFZ + 844
	9   KeyboardShortcutsExample            0x000000010218b614 $s17KeyboardShortcutsAAO8register33_4B41ED0613F74375FB61C6AA00A4208CLLyyAB8ShortcutVFZ + 224
	10  KeyboardShortcutsExample            0x000000010218c708 $s17KeyboardShortcutsAAO24registerShortcutIfNeeded33_4B41ED0613F74375FB61C6AA00A4208CLL3foryAB4NameV_tFZ + 92
	11  KeyboardShortcutsExample            0x000000010218f144 $s17KeyboardShortcutsAAO6events3forScSyAB9EventTypeOGAB4NameV_tFZyScS12ContinuationVyAF_GXEfU_ + 1952
	12  libswift_Concurrency.dylib          0x0000000230a795b0 $sScS_15bufferingPolicy_ScSyxGxm_ScS12ContinuationV09BufferingB0Oyx__GyADyx_GXEtcfC + 252
	13  KeyboardShortcutsExample            0x000000010218e990 $s17KeyboardShortcutsAAO6events3forScSyAB9EventTypeOGAB4NameV_tFZ + 272
	14  KeyboardShortcutsExample            0x00000001021a8ddc $s7SwiftUI4ViewP17KeyboardShortcutsE02onD8Shortcut_4type7performQrA2DO4NameV_AH9EventTypeOyyctFyyYaYbcfU_ + 276
	15  KeyboardShortcutsExample            0x00000001021a9189 $s7SwiftUI4ViewP17KeyboardShortcutsE02onD8Shortcut_4type7performQrA2DO4NameV_AH9EventTypeOyyctFyyYaYbcfU_TATQ0_ + 1
	16  SwiftUI                             0x00000001d66ade95 $s7SwiftUI13_TaskModifierV05InnerD033_293A0AF83C78DECE53AFAAF3EDCBA9D4LLV4body7contentQrAA05_ViewD8_ContentVyAFG_tFyycfU_yyYaYbcfU_TQ0_ + 1
	17  SwiftUI                             0x00000001d66c8395 $s7SwiftUI13_TaskModifierV05InnerD033_293A0AF83C78DECE53AFAAF3EDCBA9D4LLV4body7contentQrAA05_ViewD8_ContentVyAFG_tFyycfU_yyYaYbcfU_TATQ0_ + 1
	18  SwiftUI                             0x00000001d66c2255 $sxIeghHr_xs5Error_pIegHrzo_s8SendableRzs5NeverORs_r0_lTRyt_Tg5TQ0_ + 1
	19  SwiftUI                             0x00000001d66c6221 $sxIeghHr_xs5Error_pIegHrzo_s8SendableRzs5NeverORs_r0_lTRyt_Tg5TATQ0_ + 1
	20  libswift_Concurrency.dylib          0x0000000230a95f09 _ZL23completeTaskWithClosurePN5swift12AsyncContextEPNS_10SwiftErrorE + 1
)
```

</details>

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -407,6 +407,30 @@ extension KeyboardShortcuts {
...
+				streamKeyDownHandlers[name]?[id] = nil
+				streamKeyUpHandlers[name]?[id] = nil
+
+				unregisterShortcutIfNeeded(for: name)
+			}
+		}
+	}
+
+	@available(macOS 10.15, *)
+	@available(*, deprecated, message: "Use events(for:) instead")
```
sindresorhus, line 433: Use the `renamed:` parameter. And make it rename to the `type` overload `events()` method.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -407,6 +407,30 @@ extension KeyboardShortcuts {
 	- Note: This method is not affected by `.removeAllHandlers()`.
 	*/
 	@available(macOS 10.15, *)
+	public static func events(for name: Name) -> AsyncStream<KeyboardShortcuts.EventType> {
```
sindresorhus, line 410: You also need to add the overload with the `type` parameter, which is what most users will want to use, as you usually only need one of the events. And this makes the common case slightly more convenient.

sindresorhus: The overload could simply do something like: `events(for: name).filter { $0 == type }`

Sources/KeyboardShortcuts/ViewModifiers.swift
```diff
@@ -3,25 +3,18 @@ import SwiftUI
...
-					perform(.keyUp)
-				}
+		task {
+			for await event in KeyboardShortcuts.events(for: shortcut) {
+				perform(event)
 			}
+		}
 	}
 
 	public func onKeyboardShortcut(_ shortcut: KeyboardShortcuts.Name, event: KeyboardShortcuts.EventType, perform: @escaping () -> Void) -> some View {
```
sindresorhus: Needs a doc comment

Sources/KeyboardShortcuts/ViewModifiers.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftUI
+
+@available(macOS 12, *)
+extension View {
+	public func onKeyboardShortcut(_ shortcut: KeyboardShortcuts.Name, perform: @escaping (KeyboardShortcuts.EventType) -> Void) -> some View {
+		task {
+			for await event in KeyboardShortcuts.events(for: shortcut) {
```
sindresorhus, line 7:
```suggestion
			for await eventType in KeyboardShortcuts.events(for: shortcut) {
```

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -407,6 +407,30 @@ extension KeyboardShortcuts {
...
+
+				unregisterShortcutIfNeeded(for: name)
+			}
+		}
+	}
+
+	@available(macOS 10.15, *)
+	@available(*, deprecated, message: "Use events(for:) instead")
 	public static func on(_ type: EventType, for name: Name) -> AsyncStream<Void> {
 		AsyncStream { continuation in
```
sindresorhus, line 435: Can you make the `on` method use `events()` internally? To reduce the amount of duplicated code.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/ViewModifiers.swift
sindresorhus: > I also ran into a crash that I think is unrelated (it happens on the new events function and on the existing on function). I don't know if it's something you've seen before:

I have not seen that. How do I reproduce? Did it happen during a test run or usage?

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
EmilioPelaez: Done. I named the replacement `events(_ type: EventType, for name: Name)`, which keeps the parameter order.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
EmilioPelaez: I did this with the function `public static func events(_ type: EventType, for name: Name) -> AsyncFilterSequence<AsyncStream<EventType>>`, but I was unable to find a way to keep the return type to `AsyncStream<EventType>`.

`some AsyncSequence` exits but it erases the return value and needs to be called with `try`.

I looked at the evolution proposals and Swift forums but was unable to find a satisfactory alternative.

sindresorhus: Yeah, we need `some AsyncSequence<Element>` (hopefully coming in Swift 5.7) and rethrowing protocols.

EmilioPelaez: Alright, I'll leave it like this for now then

Sources/KeyboardShortcuts/ViewModifiers.swift
EmilioPelaez: I'm seeing this quite consistently when running from Xcode. It doesn't happen when there are not shortcuts set yet.

EmilioPelaez: @sindresorhus It seems like it was a thread issue. I wrapped the event code in a dispatch to the main thread, and it no longer crashes.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -411,20 +411,24 @@ extension KeyboardShortcuts {
...
 			let id = UUID()
 
-			streamKeyDownHandlers[name, default: [:]][id] = {
-				continuation.yield(.keyDown)
-			}
-			streamKeyUpHandlers[name, default: [:]][id] = {
-				continuation.yield(.keyUp)
-			}
+			DispatchQueue.main.async {
```
sindresorhus, line 414: Would be better to use `@MainActor`.

EmilioPelaez: I tried using `@MainActor` but I got a bunch or errors saying `Static property 'streamKeyDownHandlers' isolated to global actor 'MainActor' can not be mutated from a non-isolated context`.

To be honest, I don't understand actors well enough to know how to implement it that way.

sindresorhus: The error is protecting you from a possible race condition.

But yes, probably not worth caring about right now.

sindresorhus:
Looks good. Thank you 👍

PR #82, opened 2022-05-03
Fix compatibility with `View#focused()`

### LINKED ISSUE #76: can't focus on KeyboardShortcuts Recorder input
```swift
...
@FocusState private var shouldRecordCustomShortcut: Bool
...
.onChange(of: enterNormalModeWith) { _ in
    shouldRecordCustomShortcut = true
}
...
KeyboardShortcuts.Recorder(for: .enterNormalMode).focused($shouldRecordCustomShortcut)
```

i would expect the `KeyboardShortcuts Recorder` to be focused and being able to type a shortcut to be recorded. but although the code removes the focus from any other field, the `KeyboardShortcuts Recorder` doesn't get focused.

if i replace the `KeyboardShortcuts Recorder` by any other field, it works. so my code itself seems to be fine.

any more info i could give to help? thanks.

sindresorhus:
Fixes #76

@godbout The problem was that the recorder needs to prevent initial focus, otherwise it would enter the key recording when the window shows. I previously solved this by preventing it from becoming a key view. That's what prevented focus.

I'm not really happy with what's in this pull request, but it's the best I could come up with. I'm open to other ideas. I'm also concerned it may break some usage, although it worked fine in the example app and my Dato app.

You also need a dispatch call on before setting the focus. I think that's a SwiftUI bug and unrelated to this.

Could you try out this pull request in your app?

### DIFF
--- a/Example/KeyboardShortcutsExample.xcodeproj/project.pbxproj
+++ b/Example/KeyboardShortcutsExample.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 54;
+	objectVersion = 55;
 	objects = {
 
 /* Begin PBXBuildFile section */
@@ -127,7 +127,7 @@
 				};
 			};
 			buildConfigurationList = E36FB9412609BA43004272D9 /* Build configuration list for PBXProject "KeyboardShortcutsExample" */;
-			compatibilityVersion = "Xcode 12.0";
+			compatibilityVersion = "Xcode 13.0";
 			developmentRegion = en;
 			hasScannedForEncodings = 0;
 			knownRegions = (
@@ -222,7 +222,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				MACOSX_DEPLOYMENT_TARGET = 12.0;
+				MACOSX_DEPLOYMENT_TARGET = 12.3;
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
 				ONLY_ACTIVE_ARCH = YES;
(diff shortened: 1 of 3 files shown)


godbout:
ah! thank you. and thanks for the explanation, always eager to know what's going on. no worries about the SwiftUI bug, i've been dealing with those for the past 2 years 😅️ i know the pain.

i'll pull tomorrow or the day after and will keep you updated.

godbout:
https://user-images.githubusercontent.com/121373/166643382-5f9e23c3-1868-4103-95e4-c11276d5edd4.mp4

works beautifully man, thanks. using the dispatch call on my Picker, the focus for the `sequence` works fine too, so no side effect. seems very good to me.

godbout:
> I'm not really happy with what's in this pull request, but it's the best I could come up with. I'm open to other ideas.

wish i could help but currently absolutely no time to dig into this, sorry.

> I'm also concerned it may break some usage, although it worked fine in the example app and my Dato app.

as far as i'm concerned, it's 👍️
up to you to decide if it's ok to merge it or not, but if you do i'll take it!

sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/v1.7.0

You do not need the `DispatchQueue.main.async` call anymore.

godbout:
thank you!

godbout:
> You do not need the `DispatchQueue.main.async` call anymore.

FWIW in my case i still need the Dispatch else it fails sometimes.

godbout:
with Dispatch:

https://user-images.githubusercontent.com/121373/172125995-1927aa4a-07db-44b2-8b28-53860344d574.mp4

without Dispatch:

https://user-images.githubusercontent.com/121373/172126126-ab900172-bfdb-4b89-b555-0fe9cb006f6d.mp4

using the Dispatch makes it work all good for me, so no worries. thank you!

sindresorhus:
Then it might be an issue with SwiftUI. There are some issues with focus. I cannot reproduce it with the example app.

godbout:
no worries. i might be doing something wrong on my side. not very SwiftUI-proficient here. (although the focus is quite straightforward.)

PR #86, opened 2022-05-30
Configuration callback property to adjust AppKit object

Maschina:
I have often the need to tune AppKit objects being embedded into SwiftUI with NSViewRepresentable. This configuration property is a proposal for an optional parameter to directly access the RecorderCocoa object (similar to [introspect](https://github.com/siteline/SwiftUI-Introspect))

### DIFF
--- a/Sources/KeyboardShortcuts/Recorder.swift
+++ b/Sources/KeyboardShortcuts/Recorder.swift
@@ -7,13 +7,18 @@ extension KeyboardShortcuts {
 
 		let name: Name
 		let onChange: ((_ shortcut: Shortcut?) -> Void)?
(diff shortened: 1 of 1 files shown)


sindresorhus:
If this would have been accepted, I think it would be more SwiftUI-like to make it a builder method: `KeyboardShortcuts.Recorder(…).configure { }`.

Maschina:
> If this would have been accepted, I think it would be more SwiftUI-like to make it a builder method: `KeyboardShortcuts.Recorder(…).configure { }`.

Makes sense. In terms of flexibility and extendability, I highly encourage to consider such an option. If you agree, I can change it towards a modifier solution.

PR #88, opened 2022-06-25
Fix duplication checking for shortcuts with a shift key.

megabitsenmzq:
Try to fix shortcuts with a shift key that can not be matched by the duplication check.

Like Command+Shift+P.
If you have a menu item using this shortcut, then you tried to set the same shortcut in KeyboardShorts. The duplication check won't show any alert. Because:

For the shortcut you want to set:
keyToCharacter() = "p"
modifiers = [.shift, .command]

For the menu item:
keyEquivalent = "P"
keyEquivalentModifierMask = [.command]

They represent this shortcut in a different way. So I did this quick fix. Hope it helps.

### DIFF
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -112,12 +112,20 @@ extension KeyboardShortcuts.Shortcut {
 	*/
 	func menuItemWithMatchingShortcut(in menu: NSMenu) -> NSMenuItem? {
 		for item in menu.items {
-			if
-				keyToCharacter() == item.keyEquivalent,
-				modifiers == item.keyEquivalentModifierMask
-			{
-				return item
-			}
+            var keyEquivalent = item.keyEquivalent
+            var keyEquivalentModifierMask = item.keyEquivalentModifierMask
+            
+            if modifiers.contains(.shift) {
+                keyEquivalent = keyEquivalent.lowercased()
+                keyEquivalentModifierMask.insert(.shift)
+            }
+            
+            if
+                keyToCharacter() == keyEquivalent,
+                modifiers == keyEquivalentModifierMask
+            {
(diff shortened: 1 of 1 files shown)


sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/v1.7.1

francisfeng:
I think this fix is blocking some keyboard shortcuts with Shift as a modifier.

Say I have a Copy item in main menu with **⌘ + C**, and I tried to set **⌘ + ⇧ + C** as global keyboard shortcut for some action. A alert would pop out. Please see the screenshot for reference.

If I remove this piece of code, I can set **⌘ + ⇧ + C** as the keyboard shortcut again.

CleanShot 2022-08-16 at 14 41 32@2x

megabitsenmzq:
> I think this fix is blocking some keyboard shortcuts with Shift as a modifier.
> 
> Say I have a Copy item in main menu with **⌘ + C**, and I tried to set **⌘ + ⇧ + C** as global keyboard shortcut for some action. A alert would pop out. Please see the screenshot for reference.
> 
> If I remove this piece of code, I can set **⌘ + ⇧ + C** as the keyboard shortcut again.
> 
> CleanShot 2022-08-16 at 14 41 32@2x

I just tried this, you are right. 
It seems contains(.shift) return true even when the modifiers not containing a shift.
This is a little bit confuse. Do you have any ideas? 🤦‍♂️

megabitsenmzq:
Never mind. I made a mistake. It's not because of this. I will investigate this further. 😂

megabitsenmzq:
> Never mind. I made a mistake. It's not because of this. I will investigate this further. 😂

Got it. This is my fault. Add ", keyEquivalent.lowercased() != keyEquivalent" after "if modifiers.contains(.shift)," can fix this. I will create a pull request later.

PR #92, opened 2022-08-01
Registering / Enabling shortcuts now returns a @discardable Bool to know whether it worked or not

eternalstorms:
Also adds an "isEnabled" calculated property to Shortcut, and the appropriate .shortcutIsEnabled(_ : ) function to KeyboardShortcuts.

Makes var isTakenBySystem public so it can be used for "default" shortcuts created by an app, which makes us able to provide alternatives on startup.

This is non-breaking.

===
I squashed the 4 local commits beforehand, but I'm not sure if it worked, this is my first pull request on GitHub. It still shows me 4 commits instead of one.

### DIFF
--- a/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
@@ -37,34 +37,32 @@ enum CarbonKeyboardShortcuts {
 	private static var hotKeyId = 0
 	private static var eventHandler: EventHandlerRef?
 
-	private static func setUpEventHandlerIfNeeded() {
-		guard
-			eventHandler == nil,
-			let dispatcher = GetEventDispatcherTarget()
-		else {
-			return
-		}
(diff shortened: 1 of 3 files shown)


eternalstorms:
closed because I effed up the squashing. retrying.

Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
```diff
@@ -37,34 +37,32 @@ enum CarbonKeyboardShortcuts {
...
-		InstallEventHandler(
+		guard InstallEventHandler(
 			dispatcher,
 			carbonKeyboardShortcutsEventHandler,
 			eventSpecs.count,
 			eventSpecs,
 			nil,
 			&eventHandler
-		)
+		) == noErr else { return false }
```
sindresorhus, line 56: I think this should throw an error instead. More Swifty. Same with `GetEventDispatcherTarget`. Then this function doesn't need to return a bool.

PR #93, opened 2022-08-02
Fix layout in NSPopover

### LINKED ISSUE #75: placing `Recorder` in `NSPopover` translates popover out of position
Default behavior for my application:
Screen Shot 2022-04-05 at 11 13 59 PM
Here is the affiliated release
https://github.com/bnidevs/Captura/releases/tag/V1.7

Behavior after adding `Recorder` object:
Screen Shot 2022-04-05 at 11 14 19 PM
Here is the affiliated release
https://github.com/bnidevs/Captura/releases/tag/V1.8

I have a hunch this has something to do with the attributes of the `Recorder` object but I cannot confirm

robb:
This maintains the `minimumWidth` by overriding `sizeThatFits(:)` instead of using Auto Layout. It appears that one of the implicit constraints disabled by `self.translatesAutoresizingMaskIntoConstraints` is critical for the `NSPopover` to maintain its position.

Fixes #75, https://github.com/bnidevs/Captura/issues/2.

### DIFF
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -65,6 +65,13 @@ extension KeyboardShortcuts {
 			return size
 		}
 
+		public override func sizeThatFits(_ size: NSSize) -> NSSize {
+			var result = size
+			result.width = max(minimumWidth, result.width)
+
+			return result
+		}
+
 		private var cancelButton: NSButtonCell?
 
 		private var showsCancelButton: Bool {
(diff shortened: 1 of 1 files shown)


sindresorhus:
From the sizeThatFits docs:

> By default, this method returns the intrinsicContentSize of the receiver.

So it seems to me it's not needed at all.

sindresorhus:
Friendly bump :)

robb:
ah, apologies this fell off my radar – appreciate you landing it on your own tho!

PR #95, opened 2022-08-20
Added French and Dutch localisation

DeLoeribas:


sindresorhus:
Do you know anyone that could review the translations?

sindresorhus:
Can you change from "preferences" to "settings" as that's what it's called in macOS 13. It will avoid having to change it in the future.

Issue #97, opened 2022-10-14
Update Localizable.strings keyboard_shortcuts_can_be_changed for Ventura


stammy:
For users on Ventura, instructions for how to change an existing system-wide shortcut are a bit different since System Preferences has been renamed to System Settings and the Shortcuts has been modified a bit as well. 

`"keyboard_shortcuts_can_be_changed" = "Most system-wide keyboard shortcuts can be changed in “System Preferences › Keyboard › Shortcuts”.";
`

should become:

`"keyboard_shortcuts_can_be_changed" = "Most system-wide keyboard shortcuts can be changed in “System Settings › Keyboard ›  Keyboard Shortcuts…”.";
`

sindresorhus:
English was updated in v1.8.0.

I'll need other to handle the localisations.

sindresorhus:
// @YazeedAlKhalaf @isametry @Maschina @kant @fejese @2Grey @L1cardo @francisfeng

isametry:
I've changed the word in Czech and opened a PR.

isametry:
By the way, there is _no_ name change in German (de), since there it has been always called an equivalent of "settings" _(Einstellungen)_ since the beginning of OS X. So the German localisation can be already considered current.

FYI, a quick and simple source is the official Ventura changelog: https://www.apple.com/de/macos/ventura/features/. For another language, simply change the "de" to another abbreviation in the URL.

sindresorhus:
Note that you also need update the last part from `Shortcuts` to `Keyboard Shortcuts`.

2Grey:
Does it make sense to add a new localization string for macOS 13.0+.

Because, otherwise the current localized text will become irrelevant for older versions of macOS.

sindresorhus:
> Does it make sense to add a new localization string for macOS 13.0+?

For simplicity, I have decided to just switch to the macOS 13 strings. This text is not super important and the user should be able to figure it out regardless.

2Grey:
Ok, got you.

Also there isn't need to change Russian localization, because for Russian language name of menu items didn't change.

fejese:
I'm happy to take a look at this but I don't have Ventura yet to confirm how the Hungarian menu changed

fejese:
finally got around to check: Hungarian translation is up to date too

PR #99, opened 2022-10-20
Update `cs` localization for macOS 13

isametry:
- Changed "System Preferences" to "System Settings"
- Added a missing quotation mark


sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/issues/97#issuecomment-1285459836

PR #103, opened 2022-10-20
Updated Spanish localizations for macOS 13 (#113)

kant:


sindresorhus:
You need to change "Preferencias del sistema" too.

PR #107, opened 2022-10-27
Update KeyboardShortcuts.swift

iSapozhnik:
I guess that was the intent - allow the outside control whether handlers are called or not.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -35,7 +35,7 @@ public enum KeyboardShortcuts {
 	/**
 	When `true`, event handlers will not be called for registered keyboard shortcuts.
 	*/
-	static var isPaused = false
+	public static var isPaused = false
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
No that was not the intent. That's why it's not public. I'm happy to consider exposing it if you share a use-case.

iSapozhnik:
The use case is pretty simple - my app supports custom user-defined shortcuts. The app has a trial period. During this period, the user can make shortcuts, but when the trial expires, I wanted to pause all the shortcuts. I'm considering the usage of dynamic shortcuts as a "Pro" feature.

sindresorhus:
`isPaused` is not meant to be used for a long time. It only ignores events.

For example, in your case, the user would be expect to be able to use those keyboard shortcuts in other apps since they're not actually used in your app, but they would not be able to because your app has them registered.

I think a `.isEnabled` getter/setter (which would enable/disable all shortcuts) would be a better solution. Basically, like the `.enabled()`/`.disabled()` methods, but for all the shortcuts.

For now, you could manually use the `.enabled()`/`.disabled()` on your keyboard shortcuts.

PR #114, opened 2022-12-10
userDefaultsRemove uses removeObject

kaunteya:
At present `userDefaultsRemove` is setting the value to false.
For my specifc case, where a lot of dynamic keys are used, I noticed a lot of stale keys in UserDefaults. I think it's best to remove the key


sindresorhus:
The behavior is intentional to make the [`Name(:default)`](https://swiftpackageindex.com/sindresorhus/keyboardshortcuts/v1.9.0/documentation/keyboardshortcuts/keyboardshortcuts/name/init(_:default:)) parameter work.

kaunteya:
Ok. Is there a way to completely remove a key?

kaunteya:
I have updated `setShortcut(_ shortcut: Shortcut?, for name: Name)` method such that if the name has `defaultKey` then it will set the key to false else it will remove it.

I have also added a method `userDefaultsDisable` and updated the behaviour of `userDefaultsRemove`.
Please check the updated code

sindresorhus:
That seems like a good compromise 👍

PR #122, opened 2023-02-15
Add NSMenu support

### LINKED ISSUE #1: Support menu bar apps that uses `NSMenu`

### LINKED ISSUE #108: Add `.isEnabled` getter/setter
To temporarily disable/enable all keyboard shortcuts.

See: https://github.com/sindresorhus/KeyboardShortcuts/pull/107#issuecomment-1295829557

decodism:
Fixes #1 
Fixes #108

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -37,6 +37,35 @@ public enum KeyboardShortcuts {
 	*/
 	static var isPaused = false
 
+	/**
+	Set according to the opening state of your NSMenu if you want your keyboard shortcuts to work when it is open.
+
+	```swift
+	let menu = NSMenu()
+	let menuDelegate = MenuDelegate()
+	menu.delegate = menuDelegate
+
+	class MenuDelegate: NSObject, NSMenuDelegate {
+		func menuWillOpen(_ menu: NSMenu) {
+			KeyboardShortcuts.isMenuOpen = true
+		}
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
Nice work! I just tested it in one of my apps and it seems to work great.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -37,6 +37,26 @@ public enum KeyboardShortcuts {
...
+		didSet {
+			guard isEnabled != oldValue else {
+				return
+			}
+
+			CarbonKeyboardShortcuts.setUpRelevantEventHandler()
+		}
+	}
+
+	public static var isMenuOpen = false {
```
sindresorhus, line 50: Would you be able to add doc comments to these properties?

You can use https://chat.openai.com to help out with the text.

PR #124, opened 2023-03-04
Fix: cursor disappear when switching to other textfield

nathanwhy:
When there are RecorderCocoa and TextFields in the same view controller, cursor disappears when switching from RecorderCocoa  to other TextFields.
It seems that only RecorderCocoa has this issue, SwiftUI works well.


sindresorhus:
@nathanwhy This change causes the insertion pointer to show for a second if you press the Tab key while the recorder is focused. This is unwanted behavior.

Issue #33, opened 2020-12-04
Allow some system-reserved keyboard shortcuts that are not used?


godbout:
great package. i'm new to macOS and Swift development and got things working painlessly 👍🏼️

wondering though whether there's a way i could see why some shortcuts are said to be used system-wide while they're not registered in the macOS keyboard shortcuts settings. any idea? i.e. "Command + Esc". some other apps like 1Password give the same warning message, but some others like Dash allow the use of those shortcuts. so i guess there's an implementation difference (which is fine). just trying to understand what those system-wide shortcuts that are not registered in the macOS settings are doing and if i can change them. thanks!

sindresorhus:
The reserved keyboard shortcuts are [fetched from the system](https://github.com/sindresorhus/KeyboardShortcuts/blob/02a18776d93afef292e372bfe99d07d7c4a07805/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift#L155-L181). It doesn't provide any reason or label for why they're reserved.

Relevant Stack Overflow question: https://stackoverflow.com/questions/36044178/carbon-get-labels-for-copysymbolichotkeys

sindresorhus:
> but some others like Dash allow the use of those shortcuts

That's a bug in Dash then. This package and 1Password are doing the correct thing by respecting the system info.

godbout:
fair enough. thanks for taking the time.

Dash is not the only one. iTerm does the same. actually Dash is pretty smart about it. it doesn't allow Menu Items like copy/paste shortcuts, but it allows some others (shortcuts). `Keyboard Shortcuts` prevents me from using a bunch of shortcuts that are not in the Keyboard Shortcuts Preferences—and that actually don't seem to do anything system wise—but allows me to use some that are actually there. i'm sure you're _correct_ in the way this is supposed to be handled as per the Apple APIs or something related, just it does feel confusing as a newcomer (it's on me).

thanks again for the time, and the work.

sindresorhus:
Have you tried Googling those keyboard shortcuts? They might be for something that is not listed in system preferences and that is not immediately obvious. Maybe accessibility or something. However, if we could prove that it's actually completely unused, we could add some kind of hard-coded allow list for such keyboard shortcuts to allow them. I try to be practical about this.

Could you list the shortcuts? I'm going to try some other keyboard shortcuts solutions and see how they handle it.

godbout:
[...]

PR #125, opened 2023-03-11
Add buttons to alert on shortcut taken by system to allow user to use it anyway

sohocoke:
Fixes issue #33 (assuming it didn't cover the 'used by menu item' case)

### DIFF
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -248,19 +248,27 @@ extension KeyboardShortcuts {
 					return nil
 				}
 
-				guard !shortcut.isTakenBySystem else {
+				if shortcut.isTakenBySystem {
 					self.blur()
 
-					NSAlert.showModal(
+					let modalResponse = NSAlert.showModal(
 						for: self.window,
 						title: "keyboard_shortcut_used_by_system".localized,
 						// TODO: Add button to offer to open the relevant system settings pane for the user.
-						message: "keyboard_shortcuts_can_be_changed".localized
+						message: "keyboard_shortcuts_can_be_changed".localized,
+						buttonTitles: [
+							"force_use_shortcut".localized,
+							"Cancel"
+						]
 					)
 
 					self.focus()
 
-					return nil
+					// first button returned from alert means user wants to continue setting this 
+					// shortcut despite it being used by system.
+					guard modalResponse == .alertFirstButtonReturn else {
+						return nil
+					}
 				}
 
 				self.stringValue = "\(shortcut)"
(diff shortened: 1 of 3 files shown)


romadryud:
@sindresorhus Could you please merge this one?

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -248,19 +248,27 @@ extension KeyboardShortcuts {
...
 					self.blur()
 
-					NSAlert.showModal(
+					let modalResponse = NSAlert.showModal(
 						for: self.window,
 						title: "keyboard_shortcut_used_by_system".localized,
 						// TODO: Add button to offer to open the relevant system settings pane for the user.
-						message: "keyboard_shortcuts_can_be_changed".localized
+						message: "keyboard_shortcuts_can_be_changed".localized,
+						buttonTitles: [
```
sindresorhus, line 259: The first item should be `OK`, the second item should be `Use Anyway`, since the default choice should be `OK`.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/RecorderCocoa.swift
sohocoke: Hi @sindresorhus, I'm a great fan of your work. Using many of the projects you released for the apps I'm working on – thanks.

I had a quick look at the Apple HIG on this, and this example on alerts is the closest I could get to. 
Screenshot 2023-04-04 at 22 23 22

So things boil down to what the primary choice would be in this situation.

My rationale for the order of the buttons, which I think fits with what's outlined in the HIG, is: a 'cancel' action at the left (or bottom now?) for the user to go back and re-take an action, or a 'go ahead' action at the right (or top?) for her to proceed with the choice, having digested the message in the dialog. I.e. 'go ahead' should be the primary action.

So I think that, even if the action were to be interpreted as strongly inadvisable, the button to go ahead should be the first one. Else, the default choice would be to cancel the action, which I can't remember to be the order of choices in alerts I've seen. 

I'm happy to be proven wrong with any example alert available in a recent version of macOS though!

sohocoke:
> 

Let me know where I can see these lint errors and I'd be happy to fix them.

Sources/KeyboardShortcuts/RecorderCocoa.swift
sindresorhus: In this case, the primary button should be "OK." The primary button should represent the default, recommended, or most likely action that users would take. Since "Use Anyway" is not recommended, it should not be the primary button. Making "OK" the primary button will help guide users towards the safer and more desirable action. This follows user interface design best practices, which prioritize guiding users to make the right decisions and avoid potential issues.

This one is closer to ours:

Screen Shot 2018-10-10 at 09

sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/pull/125/files#diff-9e15662c1226f31638d9cb9aab477ef583dae71429e38a0ea8c7672e00108ff8R164-R165

sindresorhus:
Bump

PR #128, opened 2023-04-10
Add numeric keypad symbols in keyToCharacterMapping

soundflix:
numeric keypad keys are distinct triggers to their counterparts from main keyboard. They should be represented by different symbols to the user.
My solution uses Unicode COMBINING ENCLOSING KEYCAP.

### DIFF
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -186,7 +186,7 @@ private var keyToCharacterMapping: [KeyboardShortcuts.Key: String] = [
 	.f18: "F18",
 	.f19: "F19",
 	.f20: "F20",
-	
+
 	// Representations for numeric keypad keys with   ⃣  Unicode U+20e3 'COMBINING ENCLOSING KEYCAP'
 	.keypad0: "0\u{20e3}",
 	.keypad1: "1\u{20e3}",
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -185,7 +185,30 @@ private var keyToCharacterMapping: [KeyboardShortcuts.Key: String] = [
 	.f17: "F17",
 	.f18: "F18",
 	.f19: "F19",
-	.f20: "F20"
+	.f20: "F20",
+	
+	// Representations for numeric keypad keys with   ⃣  Unicode U+20e3 'COMBINING ENCLOSING KEYCAP'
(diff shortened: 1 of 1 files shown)


Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -185,7 +185,30 @@ private var keyToCharacterMapping: [KeyboardShortcuts.Key: String] = [
...
 	.f18: "F18",
 	.f19: "F19",
-	.f20: "F20"
+	.f20: "F20",
+	
+	// Representations for numeric keypad keys with   ⃣  Unicode U+20e3 'COMBINING ENCLOSING KEYCAP'
+	.keypad0: "0\u{20e3}",
+	.keypad1: "1\u{20e3}",
+	.keypad2: "2\u{20e3}",
+	.keypad3: "3\u{20e3}",
```
sindresorhus, line 194: A tiny problem with this is that the keycap style will not be shown in some places, like in a menu.

Screenshot 2023-04-27 at 02 18 44

sindresorhus: An alternative would be to use circled numbers: `"\u{2462}"`, but they are quite small...

Screenshot 2023-04-27 at 02 19 56


Screenshot 2023-04-27 at 02 20 06

soundflix: I saw the circled numbers, found them to be too small.
The other alternative I see is plain strings like "keypad 3".

sindresorhus: If you have a keypad, try going to System Settings => Keyboard => Keyboard Shortcuts and see what Apple uses for these.

soundflix:
Bildschirmfoto 2023-04-26 um 22 05 58

_These are my findings in ..._
### System Settings > Keyboard > Keyboard Shortcuts

**0 1 2 3 4 5 6 7 8 9 + -**
> no visible distinction to main keyboard

 = / *
> no visible distinction, but you wouldn't have them on main keyboard, as they are shifted.

**keypad Clear**
> x in a box (but looks different than the ones I saw in unicode table)

**keypad Enter**
> edgy arrow, different than Return

**keypad decimal separator** 
> , or . depeding on locale

**All keypad keys work as distinct trigger, i.e. you will have independent shortcuts with main versus keypad keys, even if the look the same in display.**
**In addition, you can use all keypad keys as trigger without any modifier!**

sindresorhus:
Thanks for investigating. We can go for this for now. 👍

PR #134, opened 2023-07-05
Observe NSMenu

### LINKED ISSUE #129: Drop the `isMenuOpen` property?
I wonder if we instead can simply listen to `NSMenu.didBeginTrackingNotification` and `NSMenu.didEndTrackingNotification`. That would be nice than forcing users to update the property manually.

decodism:
Fixes #129

Also adds partial support for macOS 14. Menus have been rewritten and a local event monitor receives keyUp but not keyDown.

### DIFF
--- a/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
@@ -46,6 +46,18 @@ enum CarbonKeyboardShortcuts {
 		EventTypeSpec(eventClass: OSType(kEventClassKeyboard), eventKind: UInt32(kEventRawKeyUp))
 	]
 
+	// `keyDown` not working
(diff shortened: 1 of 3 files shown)


sindresorhus:
> Also adds partial support for macOS 14. Menus have been rewritten and a local event monitor receives keyUp but not keyDown.

Opened an issue: https://github.com/sindresorhus/KeyboardShortcuts/issues/135

sindresorhus:
Seems to be working well. Thank you :)

sindresorhus:
I have heard that this still works for observing key events in NSMenu on macOS 14:

```swift
let menuObserver = CFRunLoopObserverCreateWithHandler(nil, CFRunLoopActivity.beforeSources.rawValue, true, 0, { (observer, activity)
in
if let event = NSApp.currentEvent { ••• }
})
CFRunLoopAddObserver(CFRunLoopGetCurrent(), menuObserver, CFRunLoopMode.commonModes)
```

PR #136, opened 2023-07-11
Support NSMenu on macOS 14

### LINKED ISSUE #135: Support `keyDown` when NSMenu is open on macOS 14
See: https://github.com/sindresorhus/KeyboardShortcuts/pull/134

decodism:
Fixes #135

After the latest beta broke `keyUp`, I also found that we could use `CFRunLoopObserver`.

### DIFF
--- a/Sources/KeyboardShortcuts/Utilities.swift
+++ b/Sources/KeyboardShortcuts/Utilities.swift
@@ -89,16 +89,18 @@ final class LocalEventMonitor {
 
 final class RunLoopLocalEventMonitor {
 	private let runLoopMode: RunLoop.Mode
-	private let events: NSEvent.EventTypeMask
 	private let callback: (NSEvent) -> NSEvent?
 	private let observer: CFRunLoopObserver
 
-	init(runLoopMode: RunLoop.Mode, events: NSEvent.EventTypeMask, callback: @escaping (NSEvent) -> NSEvent?) {
+	init(
+		events: NSEvent.EventTypeMask,
+		runLoopMode: RunLoop.Mode,
+		callback: @escaping (NSEvent) -> NSEvent?
+	) {
 		self.runLoopMode = runLoopMode
-		self.events = events
 		self.callback = callback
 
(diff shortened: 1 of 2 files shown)


Sources/KeyboardShortcuts/Utilities.swift
```diff
@@ -87,6 +87,60 @@ final class LocalEventMonitor {
...
+		runLoopMode: RunLoop.Mode,
+		callback: @escaping (NSEvent) -> NSEvent?
+	) {
+		self.runLoopMode = runLoopMode
+		self.callback = callback
+
+		self.observer = CFRunLoopObserverCreateWithHandler(nil, CFRunLoopActivity.beforeSources.rawValue, true, 0) { _, _ in
+			var eventsToHandle = [NSEvent]()
+
+			while let eventToHandle = NSApp.nextEvent(matching: .any, until: nil, inMode: .default, dequeue: true) {
```
sindresorhus, line 106: I think you can just pass `events` to `matching` here instead of checking inside the loop.

Sources/KeyboardShortcuts/Utilities.swift
```diff
@@ -87,6 +87,60 @@ final class LocalEventMonitor {
...
+				eventsToHandle.append(eventToHandle)
+			}
+
+			for eventToHandle in eventsToHandle {
```
sindresorhus, line 110: Why not just use the while loop instead of looping again?

sindresorhus (changes requested):

Sources/KeyboardShortcuts/Utilities.swift
sindresorhus: 👍 It would be good to have a short code comment about that as it was definitely not clear by just reading the code.

Sources/KeyboardShortcuts/Utilities.swift
sindresorhus: 👍 It would be good to have a short code comment about that as it was definitely not clear by just reading the code.

sindresorhus:
Thanks :)

Issue #130, opened 2023-06-05
No such module 'Carbon.HIToolbox'


bobmosh:
Hey folks,

I just added the package to my personal Xcode project via SPM and the first thing that happens when I build is `No such module 'Carbon.HIToolbox'` in the `CarbonKeyboardShortcuts.swift`.

I tried the latest version tags as well as pulling from main, butt the issue doesn't seem to be resolvable. Anyone else have the same issue?

Cheers,
Ferdi

bobmosh:
For context I am using this in an App that will also have an iOS App, so it might be connected to that, could that be?

bobmosh:
Solved it myself!

For anyone wondering, the library only builds for macOS targets.
In order to use it in a cross platform environment, you will need to use conditional exclusion like `#if os(macOS)` on any imports / usage of code from that library (use safer methods if possible though, this is just what I used in my tests).

You will also need multiple targets (I separated my App targets into iOS/iPadOS and macOS) and remove the linked library from the `Frameworks, Libraries, and Embedded Content` section of the mobile targets.

WalterZou:
@bobmosh
I also encountered this problem. I added `#if os(macOS)` to all the places where this library is used, but it still gives an error:
![image](https:github.com/sindresorhus/KeyboardShortcuts/assets/3727208/10f668fb-d3f7-4834-ad20-7d39f4a92854)
Is there any way to resolve this case without involving multiple targets?

WalterZou:
By adding the conditional statement `#if os(macOS) ... #endif` to all the code of KeyboardShortcuts, I have resolved this issue by ensuring that the source code only compiles on macOS, without the need to build multiple targets.

sindresorhus:
> You will also need multiple targets (I separated my App targets into iOS/iPadOS and macOS) and remove the linked library from the Frameworks, Libraries, and Embedded Content section of the mobile targets.

You don't need to do that. You can make it only used for a specific platform:

<img width="688" alt="Screenshot 2023-07-26 at 19 35 26" src="https://github.com/sindresorhus/KeyboardShortcuts/assets/170270/f71720c2-392e-4532-a712-d898b717a8b1">

WalterZou:
> > You will also need multiple targets (I separated my App targets into iOS/iPadOS and macOS) and remove the linked library from the Frameworks, Libraries, and Embedded Content section of the mobile targets.
> 
> You don't need to do that. You can make it only used for a specific platform:
> 
[...]

PR #138, opened 2023-07-26
Added conditional imports for macOS in multiple files 

WalterZou:
Add conditional imports for macOS compilation to fix import errors when used in multiple platform SwiftUI apps.  #130

### DIFF
--- a/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/CarbonKeyboardShortcuts.swift
@@ -1,3 +1,4 @@
+#if os(macOS)
 import Carbon.HIToolbox
 
 private func carbonKeyboardShortcutsEventHandler(eventHandlerCall: EventHandlerCallRef?, event: EventRef?, userData: UnsafeMutableRawPointer?) -> OSStatus {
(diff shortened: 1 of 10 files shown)


sindresorhus:
It doesn't make sense to wrap every file in this. You can specify yourself in Xcode that a dependency is only meant for a certain platform.

WalterZou:
@sindresorhus 
"Thank you for your reply. KeyboardShortcuts is an amazing project. However, in scenarios like #130, when developing Multiple-Platform Apps with a unified target using Swift UI, even if KeyboardShortcuts are specified for macOS in Xcode, there might still be a compilation error only on macOS. This PR can address and resolve this issue."

WalterZou:
After upgrading to Xcode 15 beta5, I noticed that there is no longer a need to add macOS compilation conditionals, and the compilation errors are gone (but Xcode 14 still has this issue). Therefore, I have closed this pull request. Thank you very much.

WalterZou:
After upgrading to Xcode 15 beta5, I noticed that there is no longer a need to add macOS compilation conditionals, and the compilation errors are gone (but Xcode 14 still has this issue). Therefore, I have closed this pull request. Thank you very much.

Issue #18, opened 2020-07-05
Custom storage


sindresorhus:
The defaults here are great and will work for most, but some users will need to store the keyboard shortcuts differently. For example, they are migrating from a different keyboard shortcut solution or they have a different preferred way of storing the shortcuts. I think we could expose a few storage-related hooks that would solve a lot of use-cases in one go.

We could add a protocol with a few required methods:

- `get`
- `set`
- `contains`
- `remove`

The default implementation of this protocol would be the current UserDefaults storage, but a user could override it like this `KeyboardShortcuts.storage = myStorage` if needed.

Thoughts? Am I over-designing this?

fredlb:
I would love this!
My use case is that I register shortcuts dynamically, and would preferably store this like I store the rest of the data (a JSON file that the user can load/save). In fact, just NOT storing it in defaults would make it better for my use case (but I agree that its a sane default behavior).

fredlb:
@sindresorhus I took a stab at this here https://github.com/fredlb/KeyboardShortcuts/pull/1
Heavy disclaimer: I'm a Swift-noob so this is likely not the abstraction you want, but it works for my case. I'm struggling a bit with understanding how to initialize `KeyboardShortcuts` unless I do through `Name`, so that's what I went for here.

An implementation example of `StorageProvider`:
```
class InMemoryStorageProvider: StorageProvider {
    
    private var storage: [String:String?] = [:]
    
    func set(_ value: String?, forKey defaultName: String) {
        self.storage.updateValue(value, forKey: defaultName)
    }

    func remove(forKey defaultName: String) {
        self.storage.removeValue(forKey: defaultName)
    }

    func get(forKey defaultName: String) -> String? {
        guard let data = self.storage[defaultName] else {
            return nil
        }
        return data
    }
    
    func getAll() -> [String: String?] {
        return storage
    }
    
}
```
I would some feedback on this, and I'll file a proper PR if we want to go down this route :)

sindresorhus:
The problem with a solution like this (that I did not think of when initially proposing it) is that in an app you may want to both use the default storage, but also have a custom storage for some shortcuts. That's actually the case for one of my apps. I have a global keyboard shortcut for toggling the app (using the default storage), but also keyboard shortcuts per item (custom storage).

sindresorhus:
[...]

PR #139, opened 2023-07-27
Custom storage for shortcuts

fredlb:
Related to https://github.com/sindresorhus/KeyboardShortcuts/issues/18, I've tried to implement a protocol to allow for custom storage implementations. It still defaults to using UserDefaults unless anything else has been specified before running `KeyboardShortcuts.initialize()`.

You can consider this WIP, since it would probably be nice to add more tests etc, and documentation with examples how to use it.


doriandevtech (commented):
Comments about additional comment and border effects covering

PR #151, opened 2023-10-26
Allow overriding system-reserved shortcuts

### LINKED ISSUE #33: Allow some system-reserved keyboard shortcuts that are not used?

tofumatt:
Fix #33.

Updated version of #125 with the linting errors fixed, and "OK" as the default option in the dialog.

Works as-expected in my app:

https://github.com/sindresorhus/KeyboardShortcuts/assets/90871/d155611d-d1b2-4209-97b0-490cd326a1b0


Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
```diff
@@ -3,3 +3,5 @@
 "keyboard_shortcut_used_by_menu_item" = "This keyboard shortcut cannot be used as it’s already used by the “%@” menu item.";
 "keyboard_shortcut_used_by_system" = "This keyboard shortcut cannot be used as it’s already a system-wide keyboard shortcut.";
 "keyboard_shortcuts_can_be_changed" = "Most system-wide keyboard shortcuts can be changed in “System Settings › Keyboard › Keyboard Shortcuts”.";
+"force_use_shortcut" = "Use Anyway";
+"ok" = "OK";
```
tofumatt, line 7: I'm not very familiar with localisations in Swift/Xcode projects, please let me know if there's a built-in default for "OK"/"yes" dialogs provided by the localisation framework; I'm happy to use that if there is 🙂

sindresorhus: There's nothing built-in.

Would be great if you could fill in "OK" for the existing localizations though: https://applelocalization.com/macos?q=OK&l=English&l=French&l=German&l=Italian&l=Japanese&l=Spanish

sindresorhus:
What keyboard combination do you use to test the dialog (to trigger it)?

Sources/KeyboardShortcuts/Localization/en.lproj/Localizable.strings
tofumatt: Cool, I've added them for all languages in the repo, based on the text in Apple's docs. Thanks! 🙂

tofumatt:
I used "Command + Escape", which is actually my usual keyboard shortcut for an app I'm transitioning to use `KeyboardShortcuts` from another library, so at first was very confused why it wouldn't work 😅

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -283,19 +283,26 @@ extension KeyboardShortcuts {
...
-				guard !shortcut.isTakenBySystem else {
+				if shortcut.isTakenBySystem {
 					self.blur()
 
-					NSAlert.showModal(
+					let modalResponse = NSAlert.showModal(
 						for: self.window,
 						title: "keyboard_shortcut_used_by_system".localized,
 						// TODO: Add button to offer to open the relevant system settings pane for the user.
-						message: "keyboard_shortcuts_can_be_changed".localized
```
chatobn, line 293: - mensaje:"@luzmary"
- [ ]

chatobn (changes requested):

PR #154, opened 2023-11-26
Fixed memory leak in case of NSMenuItem.

sonsongithub:
NSMenuItem has been leaked after calling `setShortcuts` and `KeyboardShortcuts.removeAllHandlers`.

The leak is caused by a callback function registered in the notification center that captures an NSMenuItem.
I added the code to remove NSMenuItem from notification center explicitly in case of calling `setShortcut(nil)` to get around this issue.


sindresorhus:
Thanks :)

PR #167, opened 2024-05-06
feat: add support of visionOS

Harry-KNIGHT:
I got a project who needs this library on visionOS but this platform wasn't supported on the `Package.swift`. 

I added it, compiled on Vision Pro simulator all is OK.


sindresorhus:
That's not how the `platform` field works. It's used to define minimum platform. It doesn't mean platforms not listed are not allowed.

PR #179, opened 2024-07-13
Update Localizable.strings

Svec-Tomas:
I modified the translation to fit the strings within the buttons when recording (pressing) a shortcut.

EDIT: First time doing this, let me know, please, if all is OK, if not I will try to fix whatever is needed.


sindresorhus:
Thanks :)

PR #181, opened 2024-07-17
Added support for changeable SwiftUI .keyboardShortcuts, based on the…

scornflake:
This adds support for SwiftUI menu items (they now auto update when shortcuts are changed).

Also; it won't throw up a UI when modifying the existing shortcut.  The implementation ... isn't great (it relies on comparing the instance points of NSMenuItems), but I can't find any other alternative where I can correctly identify that a SwiftUI generated NSMenuItem is related to a given shortcut.

- Added support for changeable SwiftUI .keyboardShortcuts, based on the work done by @mbenoukaiss
- Modified RecorderCocoa so that it doesn't show a dialog if we are reassigning one of our own menu items

### DIFF
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -229,7 +229,7 @@ extension KeyboardShortcuts {
 			// This doesn't always work for SwiftUI generated NSMenuItems (from .keyboardShortcut()) - so we also use existingNSMenuForUs, above
 			KeyboardShortcuts.userDefaultsDisable(name: shortcutName)
 
-			self.eventMonitor = LocalEventMonitor(events: [.keyDown, .leftMouseUp, .rightMouseUp]) { [weak self] event in
+			eventMonitor = LocalEventMonitor(events: [.keyDown, .leftMouseUp, .rightMouseUp]) { [weak self] event in
 				guard let self else {
 					return nil
 				}
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -215,7 +215,7 @@ extension KeyboardShortcuts {
 			KeyboardShortcuts.isPaused = true // The position here matters.
 			
 			if let existingShortcut = getShortcut(for: shortcutName) {
-				// Assuming we have a consisten NSMenuItem state, find the NSMenuItem instance that represents us right now
+				// Assuming we have a consistent NSMenuItem state, find the NSMenuItem instance that represents us right now
 				// This works only if SwiftUI has updated the NSMenuItem.  It seems (macOS 15) that even if the shortcut state is updated on KeyboardShortcutView, that the actual NSMenuItem isn't refreshed
 				// This means that if we change it multiple times in a row, this lookup returns nil.
 				// What seems to work is just hanging onto the last found reference...
--- a/Sources/KeyboardShortcuts/RecorderCocoa.swift
+++ b/Sources/KeyboardShortcuts/RecorderCocoa.swift
@@ -199,8 +199,8 @@ extension KeyboardShortcuts {
 			preventBecomingKey()
 		}
 
-		private var existingNSMenuForUs: NSMenuItem? = nil
-		
+		private var existingNSMenuForUs: NSMenuItem?
+
 		/// :nodoc:
 		override public func becomeFirstResponder() -> Bool {
 			let shouldBecomeFirstResponder = super.becomeFirstResponder()
@@ -233,10 +233,10 @@ extension KeyboardShortcuts {
 				guard let self else {
 					return nil
 				}
-				
+
 				let clickPoint = self.convert(event.locationInWindow, from: nil)
 				let clickMargin = 3.0
-				
+
 				if
 					event.type == .leftMouseUp || event.type == .rightMouseUp,
 					!self.bounds.insetBy(dx: -clickMargin, dy: -clickMargin).contains(clickPoint)
@@ -244,21 +244,21 @@ extension KeyboardShortcuts {
 					self.blur()
 					return event
 				}
-				
+
 				guard event.isKeyEvent else {
 					return nil
 				}
-				
+
 				if
 					event.modifiers.isEmpty,
 					event.specialKey == .tab
 				{
 					self.blur()
-					
+
 					// We intentionally bubble up the event so it can focus the next responder.
 					return event
 				}
-				
+
 				if
 					event.modifiers.isEmpty,
 					event.keyCode == kVK_Escape // TODO: Make this strongly typed.
@@ -266,7 +266,7 @@ extension KeyboardShortcuts {
 					self.blur()
 					return nil
 				}
-				
+
 				if
 					event.modifiers.isEmpty,
 					event.specialKey == .delete
@@ -276,7 +276,7 @@ extension KeyboardShortcuts {
 					self.clear()
 					return nil
 				}
-				
+
 				// The “shift” key is not allowed without other modifiers or a function key, since it doesn't actually work.
 				guard
 					!event.modifiers.subtracting(.shift).isEmpty
@@ -286,8 +286,8 @@ extension KeyboardShortcuts {
 					NSSound.beep()
 					return nil
 				}
-				
-				
+
+
 				if let menuItem = shortcut.takenByMainMenu {
 					let isOurOwnMenuInstance = existingNSMenuForUs != nil ? existingNSMenuForUs === menuItem : false
 					if !isOurOwnMenuInstance {
@@ -304,10 +304,10 @@ extension KeyboardShortcuts {
 						return nil
 					}
 				}
-				
+
 				if shortcut.isTakenBySystem {
 					self.blur()
-					
+
 					let modalResponse = NSAlert.showModal(
 						for: self.window,
 						title: "keyboard_shortcut_used_by_system".localized,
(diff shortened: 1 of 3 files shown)


Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+		KeyboardShortcutView(shortcutName: shortcutName) {
+			self
+		}
+	}
+}
+
+// Holds the state of the shortcut, and changes that state when the shortcut changes
+// This causes the related NSMenuItem to also update (yipeee)
+@available(macOS 11.0, *)
+struct KeyboardShortcutView<Content: View>: View {
```
JaviSoto, line 24: I think the idiomatic way to do this would be to use `ViewModifier` instead

JaviSoto (commented):
Nice!

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+		self.shortcutName = shortcutName
+		self.shortcut = KeyboardShortcuts.getShortcut(for: shortcutName)
+		self.content = content
+	}
+
+	@ViewBuilder
+	var shortcutBody: some View {
+		if let shortcut, let keyEquivalent = shortcut.toKeyEquivalent() {
+			content()
+					.keyboardShortcut(keyEquivalent, modifiers: shortcut.toEventModifiers())
```
sindresorhus, line 40: Would using `keyboardShortcut` here cause it to both trigger the Button in `content()` and also the KeyboardShortcuts event listener.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+
+	init(shortcutName: KeyboardShortcuts.Name, content: @escaping () -> Content) {
+		self.shortcutName = shortcutName
+		self.shortcut = KeyboardShortcuts.getShortcut(for: shortcutName)
+		self.content = content
+	}
+
+	@ViewBuilder
+	var shortcutBody: some View {
+		if let shortcut, let keyEquivalent = shortcut.toKeyEquivalent() {
```
sindresorhus, line 38: You don't need to do `if` and ruin the view `id` when it changes. Instead, use https://developer.apple.com/documentation/financekitui/addordertowalletbutton/keyboardshortcut(_:)-883ia#

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+		} else {
+			content()
+		}
+	}
+
+	var body: some View {
+		shortcutBody
+				// Called only when the shortcut is updated
+				.onReceive(NotificationCenter.default.publisher(for: .shortcutByNameDidChange)) { notification in
+					if let name = notification.userInfo?["name"] as? KeyboardShortcuts.Name, name == shortcutName {
```
sindresorhus, line 50: Use `guard`

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+						print("Shortcut \(shortcutName) updated to: \(current?.description ?? "nil")")
+						shortcut = current
+					}
+				}
+	}
+}
+
+@available(macOS 11.0, *)
+extension KeyboardShortcuts.Shortcut {
+	func toKeyEquivalent() -> KeyEquivalent? {
```
sindresorhus, line 64: There is similar code in https://github.com/sindresorhus/KeyboardShortcuts/blob/2e5f15581fefb821d4b366e57d817be8bf12aa58/Sources/KeyboardShortcuts/Shortcut.swift#L247 It should be unified.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		return KeyEquivalent(Character(String(utf16CodeUnits: nameBuffer, count: nameLength)))
+	}
+
+	func toEventModifiers() -> SwiftUI.EventModifiers {
+		var modifiers: SwiftUI.EventModifiers = []
+
+		if self.modifiers.contains(NSEvent.ModifierFlags.command) {
+			modifiers.update(with: EventModifiers.command)
```
sindresorhus, line 97: `.insert` would be better.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+		}
+		guard osStatus == noErr else {
+			NSLog("Code: 0x%04X  Status: %+i", carbonKeyCode, osStatus);
+			return nil
+		}
+
+		return KeyEquivalent(Character(String(utf16CodeUnits: nameBuffer, count: nameLength)))
+	}
+
+	func toEventModifiers() -> SwiftUI.EventModifiers {
```
sindresorhus, line 93: Could just be a getter.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
```diff
@@ -0,0 +1,123 @@
...
+
+import Foundation
+import SwiftUI
+import KeyboardShortcuts
+import Carbon
+
+// Provides a SwiftUI like wrapper func, that feels the same as the normal SwiftUI keyboardShortcut
+@available(macOS 11.0, *)
+extension View {
+	@ViewBuilder
```
sindresorhus, line 13: This is moot

Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -153,6 +153,12 @@ extension KeyboardShortcuts.Shortcut {
 	}
 }
 
+extension KeyboardShortcuts.Key {
+	public var specialCharacterMapping: String? {
+		keyToCharacterMapping[self]
+	}
+}
```
sindresorhus, line 160: ?

Sources/KeyboardShortcuts/RecorderCocoa.swift
```diff
@@ -297,21 +318,21 @@ extension KeyboardShortcuts {
...
-
+				
 				self.stringValue = "\(shortcut)"
 				self.showsCancelButton = true
-
+				
 				self.saveShortcut(shortcut)
 				self.blur()
-
+
```
sindresorhus, line 335: Don't make unrelated changes.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
scornflake: Yup - there's a key modifier directly above.  I'm only using KeyboardShortcutView as a wrapper to store state so that the view/menu changes when the shortcut changes.  If this is possible using a viewModifier function, happy to learn about it and adjust.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
scornflake: I did test for this - if I have the app active, as well as a "global" shortcut handler installed; I only get one event. I just presumed KeyboardShortcuts was being somehow smart

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
scornflake: Except that in my case it's being included by code that is also compiled before v11. Unless I misunderstand?

Sources/KeyboardShortcuts/Shortcut.swift
scornflake: Because I wanted that mapping for other uses - and since it was there, I thought why reinvent the wheel?

Sources/KeyboardShortcuts/RecorderCocoa.swift
scornflake: Sorry - I'll try not to mess up more formatting

Sources/KeyboardShortcuts/Shortcut.swift
scornflake: I'll rename it to make it more obvious. Basically I'm using it in an app to take a key and return a nice string. Since the mapping is already in place (just private) I thought to expose a method to get it.   If however keyToCharacter() were public, I wouldn't need the additional extension at all and could just remove it.

sindresorhus: That's an unrelated change and should be done in a separate pull request.

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
sindresorhus: The comment is on `@ViewBuilder`

Sources/KeyboardShortcuts/SwiftUI+Extensions.swift
sindresorhus: No, that's a `View` extension, not a `ViewModifier`.

scornflake:
Re the comment about going over code and cleaning up: anything specific?  I thought it was clean.
Any guidelines you want me to follow?

sindresorhus:
Thank you for the work here, and apologies for the slow follow-up, but the changes were just not the right way I wanted it.

In the meantime, this capability shipped as `View.globalKeyboardShortcut(_:)`.

PR #184, opened 2024-09-21
Make unregisterall() remove UserDefaults + Make it a public method

stormychel:
- unregisterall() now also removes UserDefaults entries
- made unregisterall() public since it can be useful to clean up all shortcuts from an app


sindresorhus:
I'm not sure it makes sense to expose this anymore. The recommended API is with `.events()` and it automatically unregisters when it's no longer listening to events.

stormychel:
We have some stuck hotkeys in our app that could benefit from this, but we could off course use a custom branch to deal with it,, whatever works best for you.
Thanks for making this available!

stormychel:
Do you want me to split up the PR so only "unregisterall() now also removes UserDefaults entries" can get merged?
It is a TODO in your code...

sindresorhus:
I think it makes more sense if it's called `resetAll()`. I got confused by its use-case because of its name. I assume you need this to **reset** all the shortcuts?

I'm ok with exposing it then. But it will need doc comments. And it should probably be moved to below `public static func reset(_ names: Name...) {`.

stormychel:
It depends whether you think this is useful for others. We are already on a custom branch, and also have a method "public static func unregisterUnknownShortcuts(knownShortcutNames: [KeyboardShortcuts.Name]) {
}" which removes shortcuts unknown by the app so maybe our needs are a bit more than what others would typically need so I don't know whether these are useful for everyone. Will make the changes you requested + doc comments, thanks for reviewing!

sindresorhus:
> static func unregisterUnknownShortcuts(knownShortcutNames: [KeyboardShortcuts.Name]) {

This one is indeed too niche.

stormychel:
> > static func unregisterUnknownShortcuts(knownShortcutNames: [KeyboardShortcuts.Name]) {
> 
> This one is indeed too niche.

Totally agree :)

I will clean up my PR and add docs over the weekend.

@sindresorhus if we have a shortcut that is stuck, even after doing our unregisterUnknownShortcuts() thing, is there anything we can do to get rid of it without wiping user defaults for app completely? One of our Macs has a shortcut that is stuck this way, hence our efforts to create unregisterUnknownShortcuts(), but even this does not solve it, while the method definitely works. This is only on one development machine, I am 100% sure this is from testing work branches, and tbh I'd just clear defaults at this point, but we want to make sure this does not happen to users. Any ideas on how this could happen?

stormychel:
@sindresorhus added docs & relocated method: https://github.com/sindresorhus/KeyboardShortcuts/pull/184/commits/5dda3b423b950d49b0a1ce4d1d939a7494a19a89

sindresorhus:
> I think it makes more sense if it's called resetAll().

stormychel:
> > I think it makes more sense if it's called resetAll().

Yes, I forgot, my bad, done now: https://github.com/sindresorhus/KeyboardShortcuts/pull/184/commits/0af13cf54c2ee953d0e71b5b4ed6b4a0afcaace8

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -270,6 +263,30 @@ public enum KeyboardShortcuts {
 		reset(names)
 	}
 
+	/**
+	Reset all keyboard shortcuts and remove their stored values from `UserDefaults`.
+
+	This method resets all keyboard shortcuts registered via `KeyboardShortcuts` and removes any associated data stored in `UserDefaults`.
+
+	Use this method to completely reset the state of all keyboard shortcuts in your app, including removing any saved user-defined shortcuts and default shortcuts.
```
sindresorhus, line 271: These 3 paragraphs pretty much say the same thing.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -270,6 +263,30 @@ public enum KeyboardShortcuts {
...
+
+	- Note: This action cannot be undone. All shortcuts will be removed, including any default shortcuts defined in `Name`. If you want to reset shortcuts back to their default values, consider using `.reset(_:)` instead.
```
sindresorhus, line 273: > If you want to reset shortcuts back to their default values, consider using `.reset(_:)` instead.

I don't think it makes sense that this has different behavior than `.reset()`. `resetAll` should just be `reset` for **all**.

sindresorhus (changes requested):

sindresorhus:
You have some lint violations. See the diff.

stormychel:
Nevermind, I will just keep on using my forked version, got other stuff to do than endlessly working on one simple PR.
Thanks for considering though!

PR #187, opened 2024-10-21
Improve localizations and Fix recorder not allowing multiple modifiers when option is present

francisfeng:
- Change Traditional Chinese to Simplified Chinese in zh-Hans.
- Remove unneeded space between English words and Chinese characters.
- Fix shortcut recorder not allowing other modifiers when option is present on macOS 15.


sindresorhus:
Thanks :)

Issue #192, opened 2024-11-30
Incorrect `KeyEquivalent` generated by for `globalKeyboardShortcut(_:)` with `.space` key


cweider:
When looking to adopt `globalKeyboardShortcut(_:)` (#101, introduced: 68d5d2d), I noticed that spacebar-based shortcuts looked weird and used the letter "S":

| Actual | Expected |
|--------|--------|
| <img alt="SwiftUI KeyboardShortcut - Actual" src="https://github.com/user-attachments/assets/2da20fca-971f-476a-878b-31fbd1dbedfc"> | <img alt="SwiftUI KeyboardShortcut - Expected" src="https://github.com/user-attachments/assets/798d76d5-0026-4485-b607-f043a48beba4"> | 

This is because `.toSwiftUI` is taking the first character of the string `"Space"` yielded by `keyToCharacter()`:
https://github.com/sindresorhus/KeyboardShortcuts/blob/c3c361f409b8dbe1eab186078b41c330a6a82c9a/Sources/KeyboardShortcuts/Shortcut.swift#L360-L369

`.keyEquivalent`, which is used in connection with `NSMenuItem` accounts for this by checking string length and using an alternative mapping if `keyToCharacter()` yields more than one character:
https://github.com/sindresorhus/KeyboardShortcuts/blob/c3c361f409b8dbe1eab186078b41c330a6a82c9a/Sources/KeyboardShortcuts/Shortcut.swift#L323-L338

cweider:
A quick fix to `.toSwiftUI` exists, but taking a more thorough pass in the interest of consistency would be worthwhile. A patch is forthcoming.

cweider:
As for the weird space on the trailing edge of menu items with shortcuts that use the space key, that’s an existing (strange) OS issue (macOS 14):

<img width="242" alt="NSMenu trailing space" src="https://github.com/user-attachments/assets/326d07ae-9962-4c3d-aef6-e4b0b5722879">

PR #193, opened 2024-11-30
Fix logic for `.globalKeyboardShortcut()`

cweider:
Evaluating `globalKeyboardShortcut(_:)` revealed several issues with related to special keys:

- Shortcuts using the `Key.space` would produce a `KeyEquivalent` that specified the "S" key. (cause: `first` being used on the `keyToCharacter()` return value, `"space"`)
- Shortcuts using `Key.f*` would produce a `KeyEquivalent` that specified the "F" key. (cause: same as above)
- Shortcuts using `Key.keypad*` would produce a `KeyEquivalent` that was not limited to the number pad. (cause: neither SwiftUI nor AppKit make this distinction when producing menu items)

Considering these issues, a reevaluation of special keys' handling is called for. Excepting the `Carbon` integration that drives the actual keyboard shortcuts, there are three distinct contexts where a `Shortcut` is used:

- `Recoder`: where an `NSString` is produced to go into the `TextField` displaying the shortcut's current value
- `NSMenuItem`: for specifying `keyEquivalent` and `keyEquivalentModifierMask`
- `toSwiftUI`: for creating a `SwiftUI.KeyboardShortcut`

Ideally, each of these three do their work in analogous ways. This branch accomplishes this with the following:

- `SpecialKey` enumerates the cases that require extra attention. This is drawn directly from keys used in `keyToCharacterMapping` and used to ensure that each special case is handled.
- Three mappings for `SpecialKey` are created for each of the three contexts where the handling is important: `presentableDescription`, `swiftUIKeyEquivalent`, and `appKitMenuItemKeyEquivalent`.
- In these three contexts if there is a `SpecialKey`, the defined mapping is what rules. `keyToCharacter()` is used if and only if there is no `SpecialKey` available. An assertion to this effect is added.

Fixes: #192

### DIFF
--- a/Sources/KeyboardShortcuts/Shortcut.swift
+++ b/Sources/KeyboardShortcuts/Shortcut.swift
@@ -427,7 +427,7 @@ extension SpecialKey {
 		case .escape:
 			.escape
 		case .help:
-			UnicodeScalar(NSHelpFunctionKey).map { KeyEquivalent(Character($0)) }
+			KeyEquivalent(unicodeScalarValue: NSHelpFunctionKey)
 		case .home:
 			.home
 		case .space:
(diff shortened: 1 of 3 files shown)


cweider:
For testing, this is a snippet that you can go and stick any old place: 
```
Menu("Menu", systemImage: "filemenu.and.selection") {
	Button("Shortcut 1") {
		// no-op
	}
	.globalKeyboardShortcut(.testShortcut1)

	Button("Shortcut 2") {
		// no-op
	}
	.globalKeyboardShortcut(.testShortcut2)

	Button("Shortcut 3") {
		// no-op
	}
	.globalKeyboardShortcut(.testShortcut3)

	Button("Shortcut 4") {
		// no-op
	}
	.globalKeyboardShortcut(.testShortcut4)
}
```

cweider:
| Actual | Expected |
|--------|--------|
| SwiftUI KeyboardShortcut - Actual | SwiftUI KeyboardShortcut - Expected |

cweider:
I should add that [cweider/KeyboardShortcuts@swiftui-key-equivalent](https://github.com/cweider/KeyboardShortcuts/commit/swiftui-key-equivalent) is the minimal change for SwiftUI. That this, `keyEquivalent`, and `description` each took different approaches to the similar problem motivated this patch.

Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -347,25 +719,34 @@ extension KeyboardShortcuts.Shortcut: CustomStringConvertible {
...
+		}
+		return modifiers.presentableDescription + String(keyToCharacter() ?? "�").capitalized
+	}
+
 	@MainActor
 	public var description: String {
-		// We use `.capitalized` so it correctly handles “⌘Space”.
-		modifiers.presentableDescription + (keyToCharacter()?.capitalized ?? "�")
+		// TODO: `description` needs to be `nonisolated`
+		presentableDescription
```
cweider, line 734: Anticipating `SWIFT_STRICT_CONCURRENCY` here.

Solution would seem to be switching the UI to use `presentableDescription` while also changing to `CustomDebugStringConvertible` and having `debugDescription` avoid `keyToCharacter()` by handling `kVK_ANSI_*` directly? It doesn’t look like there’s any way around compromising the API.

Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -179,113 +179,487 @@ extension KeyboardShortcuts.Shortcut {
...
+		case .upArrow:
+			.upArrow
+		case .rightArrow:
+			.rightArrow
+		case .downArrow:
+			.downArrow
+		case .leftArrow:
+			.leftArrow
+		case .f1:
+			UnicodeScalar(NSF1FunctionKey).map { KeyEquivalent(Character($0)) }
```
sindresorhus, line 450: I would make a tiny helper function for this, to avoid all the repetition. Can be used for `case .help:` too.

Sources/KeyboardShortcuts/Shortcut.swift
```diff
@@ -179,113 +179,487 @@ extension KeyboardShortcuts.Shortcut {
...
+		case .upArrow:
+			"↑"
+		case .rightArrow:
+			"→"
+		case .downArrow:
+			"↓"
+		case .leftArrow:
+			"←"
+		case .f1:
+			UnicodeScalar(NSF1FunctionKey).map { Character($0) }
```
sindresorhus, line 565: Same as https://github.com/sindresorhus/KeyboardShortcuts/pull/193/files#r1868297646

sindresorhus:
This is looking very good. Thanks for working fixing this 🙏

Sources/KeyboardShortcuts/Shortcut.swift
sindresorhus: Maybe open an issue about this? I don't intend to enable strict concurrency anytime soon.

PR #194, opened 2024-11-30
Add methods for removing handlers for specific `Name`s

joshnies:
Adds the following methods for removing only the handlers for specific `Name`s:

```swift
/**
Remove all handlers for one or more names. Includes both "key down" and "key up" handlers.
*/
public static func removeHandlers(for names: [Name]) {}

/**
Remove all "key down" handlers for one or more names.
*/
public static func removeKeyDownHandlers(for names: [Name]) {}

/**
Remove all "key up" handlers for one or more names.
*/
public static func removeKeyUpHandlers(for names: [Name]) {}
```

> NOTE: I also made `legacyKeyDownHandlers` and `legacyKeyUpHandlers` internal for testing purposes.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -7,8 +7,8 @@ Global keyboard shortcuts for your macOS app.
 public enum KeyboardShortcuts {
 	private static var registeredShortcuts = Set<Shortcut>()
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
I'm not sure it makes sense to add these methods as `onKeyDown`/`onKeyUp` are legacy and will be deprecated in the future. The recommended API is `.events()` which does not need this.

PR #196, opened 2024-12-28
Made sure multi-platform project compiles

dominikmayer:
Removed the `KeyEquivalent` extension from iOS builds so they compile.


sindresorhus:
`KeyEquivalent` is available on iOS.

dominikmayer:
Fair point, but then you'd need to move `import SwiftUI` out of the macOS-only block and add `@available(iOS 14.0, *)` to the function.

Or does the current version compile for you?

dominikmayer:
I changed it.

PR #201, opened 2025-03-05
Expose keyEquivalent

gpoitch:
Makes `Shortcut.keyEquivalent` public.

Use case:
Say I have a menu item for "New Post". Users can set a preferred global shortcut, otherwise it uses a local-only default. I need to set the menu item's keyEquivalent/keyEquivalentModifierMask for both scenarios.


sindresorhus:
Would be nice to handle this automatically, but this is good enough for now. https://github.com/sindresorhus/KeyboardShortcuts/issues/202

sindresorhus:
https://github.com/sindresorhus/KeyboardShortcuts/releases/tag/2.3.0

PR #210, opened 2025-04-12
Custom UserDefaults for 

Maschina:
I am using KeyboardShortcuts in an environment with [App Groups](https://developer.apple.com/documentation/xcode/configuring-app-groups). The shared UserDefaults shall also include the defined shortcuts. 

This PR adds the capability to define a custom UserDefaults to support App Groups and access group containers.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -37,6 +37,7 @@ public enum KeyboardShortcuts {
 
 	private static var openMenuObserver: NSObjectProtocol?
 	private static var closeMenuObserver: NSObjectProtocol?
+	private static var userDefaultsObservers = [UserDefaultsObservation]()
 	
 	public static var customDefaults: UserDefaults? {
 		get {
@@ -428,6 +429,7 @@ public enum KeyboardShortcuts {
 	*/
 	public static func onKeyDown(for name: Name, action: @escaping () -> Void) {
 		legacyKeyDownHandlers[name, default: []].append(action)
+		startObservingShortcut(for: name)
 		registerShortcutIfNeeded(for: name)
 	}
 
@@ -454,14 +456,40 @@ public enum KeyboardShortcuts {
 	*/
 	public static func onKeyUp(for name: Name, action: @escaping () -> Void) {
 		legacyKeyUpHandlers[name, default: []].append(action)
+		startObservingShortcut(for: name)
 		registerShortcutIfNeeded(for: name)
 	}
 
 	private static let userDefaultsPrefix = "KeyboardShortcuts_"
 
-	private static func userDefaultsKey(for shortcutName: Name) -> String { "\(userDefaultsPrefix)\(shortcutName.rawValue)"
+	private static func userDefaultsKey(for shortcutName: Name) -> String { 
+		"\(userDefaultsPrefix)\(shortcutName.rawValue)"
 	}
-
+	
+	/**
+	Start observing UserDefaults changes for a specific shortcut name.
+	Only starts observation if the shortcut is not already being observed.
+	*/
+	private static func startObservingShortcut(for name: Name) {
+		let key = userDefaultsKey(for: name)
+		
+		let observation = UserDefaultsObservation(
+			suite: userDefaults,
+			name: name,
+			key: key
+		) { name, value in
+			if value == nil {
+				self.unregisterShortcutIfNeeded(for: name)
+			} else {
+				self.registerShortcutIfNeeded(for: name)
+			}
+		}
+		
+		observation.start()
+		
+		userDefaultsObservers.append(observation)
+	}
+	
 	static func userDefaultsDidChange(name: Name) {
 		// TODO: Use proper UserDefaults observation instead of this.
 		NotificationCenter.default.post(name: .shortcutByNameDidChange, object: nil, userInfo: ["name": name])
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -37,6 +37,24 @@ public enum KeyboardShortcuts {
 
 	private static var openMenuObserver: NSObjectProtocol?
 	private static var closeMenuObserver: NSObjectProtocol?
+	
+	public static var customDefaults: UserDefaults? {
+		get {
+			_customDefaults
+		}
+
+		set {
+			_customDefaults = newValue
+		}
+	}
+
+	static var _customDefaults: UserDefaults? {
+		didSet {
+			userDefaults = _customDefaults ?? .standard
+		}
+	}
+
+	static var userDefaults: UserDefaults = .standard
 
 	/**
 	When `true`, event handlers will not be called for registered keyboard shortcuts.
@@ -59,7 +77,7 @@ public enum KeyboardShortcuts {
 	}
 
 	static var allNames: Set<Name> {
-		UserDefaults.standard.dictionaryRepresentation()
+		KeyboardShortcuts.userDefaults.dictionaryRepresentation()
 			.compactMap { key, _ in
 				guard key.hasPrefix(userDefaultsPrefix) else {
 					return nil
@@ -326,7 +344,7 @@ public enum KeyboardShortcuts {
 	*/
 	public static func getShortcut(for name: Name) -> Shortcut? {
 		guard
-			let data = UserDefaults.standard.string(forKey: userDefaultsKey(for: name))?.data(using: .utf8),
+			let data = KeyboardShortcuts.userDefaults.string(forKey: userDefaultsKey(for: name))?.data(using: .utf8),
 			let decoded = try? JSONDecoder().decode(Shortcut.self, from: data)
 		else {
 			return nil
@@ -459,7 +477,7 @@ public enum KeyboardShortcuts {
 		}
 
 		register(shortcut)
-		UserDefaults.standard.set(encoded, forKey: userDefaultsKey(for: name))
+		KeyboardShortcuts.userDefaults.set(encoded, forKey: userDefaultsKey(for: name))
 		userDefaultsDidChange(name: name)
 	}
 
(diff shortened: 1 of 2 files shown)


orchetect:
Being able to substitute UserDefaults suites is valuable in other respects. In an app where I use the package, for unit tests and integration tests we use a custom suite for prefs, and it would be useful to be able to use the same suite with KeyboardShortcuts.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -37,6 +37,25 @@ public enum KeyboardShortcuts {
...
+		}
+	}
+
+	static var _customDefaults: UserDefaults? {
+		didSet {
+			userDefaults = _customDefaults ?? .standard
+		}
+	}
+
+	static var userDefaults: UserDefaults = .standard
```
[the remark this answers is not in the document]
sindresorhus: And this needs a proper documentation comment.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -436,14 +456,40 @@ public enum KeyboardShortcuts {
...
+	private static func userDefaultsKey(for shortcutName: Name) -> String { 
+		"\(userDefaultsPrefix)\(shortcutName.rawValue)"
 	}
-
+	
+	/**
+	Start observing UserDefaults changes for a specific shortcut name.
+	Only starts observation if the shortcut is not already being observed.
+	*/
+	private static func startObservingShortcut(for name: Name) {
```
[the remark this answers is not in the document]
Maschina: First and foremost, the primary purpose of this additional listener was to enable communication between apps belonging to the same App Group. I have this dedicated use case where the window app instance sets the KeyboardShortcuts, whereas another dedicated menu instance listens to keyboard events. Without this additional listener, the menu instance will never notice a change of UserDefaults while running, unless it listens via the rather expensive NotificationCenter approach (`didChangeNotification`). 

Second, I thought that this additional listener could help to migrate to a proper UserDefaults observation instead of `shortcutByNameDidChange`?

What do you think?

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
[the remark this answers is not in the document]
Maschina: Agree! Will fix it.

Sources/KeyboardShortcuts/Utilities.swift
```diff
@@ -559,3 +559,104 @@ extension Character {
...
+		
+		suite?.addObserver(
+			self,
+			forKeyPath: key,
+			options: [.new],
+			context: &Self.observationContext
+		)
+		isObserving = true
+		
+		lock.unlock()
```
sindresorhus, line 605: This is why most locking solution has a `withLock`. The code will deadlock if it's already observing as unlock is not called in the guard else.

```swift
extension NSLock {
	@discardableResult
	func withLock<T, E>(_ apply: () throws(E) -> T) throws(E) -> T {
		lock()

		defer {
			unlock()
		}

		return try apply()
	}
}
```

Sources/KeyboardShortcuts/Utilities.swift
```diff
@@ -559,3 +559,104 @@ extension Character {
 		self = Character(content)
 	}
 }
+
+final class UserDefaultsObservation: NSObject {
```
sindresorhus, line 563: This class does not need to know about `KeyboardShortcuts.Name`. It can be more general.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -478,13 +524,13 @@ public enum KeyboardShortcuts {
...
 
-		UserDefaults.standard.removeObject(forKey: userDefaultsKey(for: name))
+		KeyboardShortcuts.userDefaults.removeObject(forKey: userDefaultsKey(for: name))
 		unregister(shortcut)
 		userDefaultsDidChange(name: name)
 	}
 
 	static func userDefaultsContains(name: Name) -> Bool {
-		UserDefaults.standard.object(forKey: userDefaultsKey(for: name)) != nil
+		KeyboardShortcuts.userDefaults.object(forKey: userDefaultsKey(for: name)) != nil
```
sindresorhus, line 533: Ensure running `swiftlint` locally passes.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -436,14 +456,40 @@ public enum KeyboardShortcuts {
...
+			if value == nil {
+				self.unregisterShortcutIfNeeded(for: name)
+			} else {
+				self.registerShortcutIfNeeded(for: name)
+			}
+		}
+		
+		observation.start()
+		
+		userDefaultsObservers.append(observation)
```
sindresorhus, line 490: Doesn't look like observers are ever destroyed.

sindresorhus (changes requested):

Sources/KeyboardShortcuts/Utilities.swift
Maschina: Agree

Sources/KeyboardShortcuts/Utilities.swift
Maschina: Good catch, thanks. I'm using the `withLock` from the [Foundation library](https://github.com/swiftlang/swift-corelibs-foundation/pull/4736), if you don't mind.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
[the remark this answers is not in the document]
Maschina: If you don't mind, I would offer to migrate the other `userDefaultsDidChange`s, but in a different PR. So it's clean and easy to follow.

Maschina:
I am just realizing that event streams also require observations. Will make another commit for that.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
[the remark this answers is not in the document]
Maschina: While testing in my app, I have just realized that they simplified version is not concurrency / Swift 6 compatible all will throw a compiler error in Swift 6 mode. It can be easily fixed with `@preconcurrency import KeyboardShortcuts`, but I am wondering if it is more reasonable to provide a Swift 6 ready variant?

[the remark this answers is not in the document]
sindresorhus: I plan to make the whole thing `@MainActor`, so that's not a concern.

Sources/KeyboardShortcuts/Utilities.swift
```diff
@@ -559,3 +559,105 @@ extension Character {
...
+			let change
+		else {
+			return
+		}
+
+		guard keyPath == key else {
+			return
+		}
+
+		let encodedString = change[.newKey] as? String
```
sindresorhus, line 660: This should not assume `String`. It could be a `Bool` too: https://github.com/sindresorhus/KeyboardShortcuts/blob/5cce327a610d59569de0977e9fdda75931ccb7e1/Sources/KeyboardShortcuts/KeyboardShortcuts.swift#L548

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -437,11 +471,54 @@ public enum KeyboardShortcuts {
...
+	- When the shortcut is added or modified, registers the new shortcut
+	
+	- Parameter name: The name of the shortcut to observe
+	*/
+	private static func startObservingShortcutIfNeeded(for name: Name) {
+		let key = userDefaultsKey(for: name)
+
+		// check userDefaultsObservers to see if we are already observing this key
+		if let observer = userDefaultsObservers.first(where: { $0.key == key }) {
+			observer.start()
```
sindresorhus, line 504: I think this could use a better code comment because just looking at it, it's not immediate obvious why we may want to call start() on an already existing observation.

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -179,6 +204,12 @@ public enum KeyboardShortcuts {
 
 		legacyKeyDownHandlers = [:]
 		legacyKeyUpHandlers = [:]
+
+		// invalidate and remove all elements of userDefaultsObservers
```
sindresorhus, line 208: This should comment that it's "defensive" since it's already done when calling `removeAll()` (deinit).

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -59,7 +84,7 @@ public enum KeyboardShortcuts {
 	}
 
 	static var allNames: Set<Name> {
-		UserDefaults.standard.dictionaryRepresentation()
+		Self.userDefaults.dictionaryRepresentation()
```
sindresorhus, line 87: `Self.` is not needed.

sindresorhus (changes requested):

sindresorhus:
Make sure you manually test it in the example app and also a real app to ensure it doesn't introduce any new bugs. I'm a little bit nervous about large changes like this as we have no real tests.

Maschina:
@sindresorhus Quick update: This PR is currently being tested since last Summer in my [app](www.welovelights.app) without any fatal errors or crashes. There were also no reports of misbehavior or other issues from other users so far. 

Should we add another dedicated example app to properly test?

sindresorhus:
- `removeAllHandlers()` now invalidates and claers `userDefaultsObservers`, which also removes observations created for `events(for:)`, despite the doc note saying it does not affect `events(for:)`. That means stream listeners stop syncing to UserDefaults changes after `removeAllHandlers()` is called. Only remove observers tied to legacy handlers or keep observations separate.
- Changing `KeyboardShortcuts.userDefaults` only updates existing observers and does not reconcile currently registered shortcuts with the new suite. If `userDefaults` is changed after handlers exist, shortcuts from the old suite stay registered and the new suite does not apply until a change event occurs. Tis makes behavior stale.
- Observers are never removed when the last handler for a name is removed or when an async stream terminates. That leaves KVO observers alive and allows later defaults changes to re-register shortcuts even though there are no handlers, which can intercept keys unexpectedly. Consider removing observers when there are no handlers for that name.

PR #216, opened 2025-05-05
Remove handler of distinct keyboard shortcut

Maschina:
This PR adds a new function to remove the handler of a distinct keyboard shortcut. It extends to public function `removeAllHandlers()` for a more fine-granular control.

My app needs to unload a certain number of models that contain registered shortcuts. Once the module is unloaded, the shortcut should be unregistered. The `removeAllHandlers` does not give me the control to unregister only certain shortcuts.

### DIFF
--- a/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
+++ b/Sources/KeyboardShortcuts/KeyboardShortcuts.swift
@@ -181,6 +181,23 @@ public enum KeyboardShortcuts {
 		legacyKeyUpHandlers = [:]
 	}
 
+	/**
+	Remove the keyboard shortcut handler for the given name.
+	
+	This can be used to reset the handler before re-creating it to avoid having multiple handlers for the same shortcut.
+	
+	- Parameter name: The name of the keyboard shortcut to remove handlers for.
+	*/
+	public static func removeHandler(for name: Name) {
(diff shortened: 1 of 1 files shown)


sindresorhus:
Duplicate of #194

Maschina:
How about marking keyUp/keyDown related functions as deprecated?

Sources/KeyboardShortcuts/KeyboardShortcuts.swift
```diff
@@ -181,6 +181,23 @@ public enum KeyboardShortcuts {
 		legacyKeyUpHandlers = [:]
 	}
 
+	/**
+	Remove the keyboard shortcut handler for the given name.
+	
+	This can be used to reset the handler before re-creating it to avoid having multiple handlers for the same shortcut.
+	
+	- Parameter name: The name of the keyboard shortcut to remove handlers for.
```
sindresorhus, line 189: Needs:

```
- Note: This method does not affect listeners using `.on()`.
```

Maschina:
OK, makes sense. 
Should we plan a PR to mark the planned deprecation?

sindresorhus:
> Should we plan a PR to mark the planned deprecation?

Yeah, I plan to doc deprecate it later.

PR #226, opened 2025-09-16
feat: support multicords keybinding recording

xinnjie:
Add `ShortcutRecordView` to support multicords keybinding recording, for example, `Ctrl-K Ctrl-S`.

Note that it does not support `KeyboardShortcuts.onKeyDown` and `KeyboardShortcuts.onKeyUp` yet


xinnjie:
This could be a niche need 😅

PR #236, opened 2026-03-03
Fix option+letter shortcuts being intercepted by IME layer

### LINKED ISSUE #235: Option + letter shortcuts stop working intermittently (Carbon RegisterEventHotKey + IME conflict)
## Problem

Shortcuts using `option + letter` (e.g. `option + q`) stop working intermittently. `option + number` works reliably. Re-opening the app's settings window temporarily restores them.

This was reported in a downstream app, [FlashSpace](https://github.com/wojciech-kulik/FlashSpace/issues/558), whose author directed me here.

## Root Cause

I believe this is a fundamental conflict between Carbon's `RegisterEventHotKey` and the macOS Input Method Engine (IME).

On macOS, `option + letter` combos (e.g. `option + q` → `œ`) are processed by the Text Services Manager **before** Carbon hotkey events are dispatched. As a result:

- `option + 1` → no character composition → `kEventHotKeyPressed` fires reliably
- `option + q` → triggers IME composition (`œ`) → the event is consumed upstream and **never reaches the Carbon event handler**

This doesn't always happen — it depends on the active application's IME mode, which is why the issue appears intermittently.

## Why the Settings Window "Fixes" It

Looking at `HotKey.swift`, when a menu is open, `HotKeyCenter` switches from `RegisterEventHotKey` to a `RunLoopLocalEventMonitor` with `.eventTracking` run loop mode (raw `NSEvent` key events), which intercepts events **before** the IME layer. When the window closes, it switches back to `RegisterEventHotKey`, and the issue returns.

## Connection to `isDisallowed`

I also noticed `Shortcut.swift` has an `isDisallowed` check for `option + letter` on macOS 15.0/15.1 in sandboxed apps — which suggests this is a known Apple regression. However, `isDisallowed` only affects the `RecorderCocoa` UI, not the actual hotkey registration, so the shortcut can still be registered but may silently fail to fire.

## Question

Is there a way to use `RunLoopLocalEventMonitor` (or a `CGEventTap`) **permanently** for shortcuts that involve `option + letter`, rather than only when a menu is open? That would allow intercepting events before the IME layer at all times.

Or alternatively, could `isDisallowed` be broadened to cover more macOS versions or non-sandboxed apps, with a warning shown to users that `option + letter` shortcuts are unreliable?

## Environment

- macOS version: 26.3
- Tested via FlashSpace app

BashkaMen:
Fixes #235

## Problem

Shortcuts using `option + letter` (e.g. `option + q`) stop working intermittently. `option + number` works reliably. Re-opening the app's settings window temporarily restores them.

**Root cause:** Carbon's `RegisterEventHotKey` doesn't receive events for `option + letter` combinations because macOS processes them through the Text Services Manager / IME layer first (e.g. `option + q` → `œ`). The Carbon event never fires. This doesn't affect `option + number` because digits don't participate in IME character composition.

The existing `RunLoopLocalEventMonitor` with `.eventTracking` already solves this correctly for the menu-open case, since it intercepts raw key events before IME. This PR extends the same approach to the normal (non-menu) mode.

## Fix

When at least one registered shortcut uses `option` without `command` or `control` (the only combinations affected by IME), a `NSEvent.addGlobalMonitorForEvents` listener is activated alongside the normal Carbon hotkey handler.

The global monitor handles key matching via the existing `handleRawKeyEvent` path — no new matching logic needed. It is automatically enabled/disabled as shortcuts are registered/unregistered, and torn down entirely when no option-only shortcuts remain.

## Scope

Only `option`-only shortcuts (no `cmd`/`ctrl`) are affected. Adding `command` or `control` prevents IME interception — those shortcuts continue using Carbon exclusively.

### DIFF
--- a/Sources/KeyboardShortcuts/HotKey.swift
+++ b/Sources/KeyboardShortcuts/HotKey.swift
@@ -117,6 +117,8 @@ final class HotKeyCenter {
 		return nil
 	}
 
+	private var globalEventMonitor: AnyObject?
+
 	var mode: Mode = .normal {
 		didSet {
 			guard mode != oldValue else {
@@ -219,6 +221,7 @@ final class HotKeyCenter {
 		}
 
 		hotKeys.removeValue(forKey: hotKey.id)
+		updateEventHandler()
 	}
 
 	private func pause(_ hotKey: HotKey) {
@@ -324,6 +327,38 @@ final class HotKeyCenter {
 
 		setHotKeyEventHandlingEnabled(shouldHandleHotKeys)
 		setRawKeyEventHandlingEnabled(shouldHandleRawKeys)
+		setGlobalOptionKeyMonitorEnabled(mode == .normal && hasOptionOnlyShortcuts)
+	}
+
+	/**
+	Returns true if any registered shortcut uses Option as the only modifier (no Command or Control).
+	Such shortcuts are prone to being intercepted by the IME layer before Carbon receives them.
+	*/
+	private var hasOptionOnlyShortcuts: Bool {
+		hotKeys.values.compactMap(\.value).contains { hotKey in
+			let modifiers = NSEvent.ModifierFlags(carbon: hotKey.carbonModifiers)
+			return modifiers.contains(.option)
+				&& !modifiers.contains(.command)
+				&& !modifiers.contains(.control)
+		}
+	}
+
+	private func setGlobalOptionKeyMonitorEnabled(_ isEnabled: Bool) {
+		if isEnabled, globalEventMonitor == nil {
+			globalEventMonitor = NSEvent.addGlobalMonitorForEvents(matching: [.keyDown, .keyUp]) { [weak self] event in
+				guard
+					let self,
+					let eventRef = OpaquePointer(event.eventRef)
+				else {
+					return
+				}
+
+				_ = handleRawKeyEvent(eventRef)
+			}
+		} else if !isEnabled, let monitor = globalEventMonitor {
+			NSEvent.removeMonitor(monitor)
+			globalEventMonitor = nil
+		}
 	}
 
 	private func setHotKeyEventHandlingEnabled(_ isEnabled: Bool) {
--- a/Sources/KeyboardShortcuts/HotKey.swift
+++ b/Sources/KeyboardShortcuts/HotKey.swift
@@ -337,9 +337,7 @@ final class HotKeyCenter {
 	private var hasOptionOnlyShortcuts: Bool {
 		hotKeys.values.compactMap(\.value).contains { hotKey in
 			let modifiers = NSEvent.ModifierFlags(carbon: hotKey.carbonModifiers)
-			return modifiers.contains(.option)
-				&& !modifiers.contains(.command)
-				&& !modifiers.contains(.control)
+			return modifiers == [.option]
 		}
 	}
 


Sources/KeyboardShortcuts/HotKey.swift
```diff
@@ -324,6 +327,38 @@ final class HotKeyCenter {
...
+	Such shortcuts are prone to being intercepted by the IME layer before Carbon receives them.
+	*/
+	private var hasOptionOnlyShortcuts: Bool {
+		hotKeys.values.compactMap(\.value).contains { hotKey in
+			let modifiers = NSEvent.ModifierFlags(carbon: hotKey.carbonModifiers)
+			return modifiers.contains(.option)
+				&& !modifiers.contains(.command)
+				&& !modifiers.contains(.control)
+		}
+	}
```
wojciech-kulik, line 342: It won't work if a user has also shortcuts with different modifiers. It would be best to combine both the current solution and the new one, I guess.

Sources/KeyboardShortcuts/HotKey.swift
```diff
@@ -324,6 +327,38 @@ final class HotKeyCenter {
...
+	/**
+	Returns true if any registered shortcut uses Option as the only modifier (no Command or Control).
+	Such shortcuts are prone to being intercepted by the IME layer before Carbon receives them.
+	*/
+	private var hasOptionOnlyShortcuts: Bool {
+		hotKeys.values.compactMap(\.value).contains { hotKey in
+			let modifiers = NSEvent.ModifierFlags(carbon: hotKey.carbonModifiers)
+			return modifiers.contains(.option)
+				&& !modifiers.contains(.command)
+				&& !modifiers.contains(.control)
```
wojciech-kulik: Probably a more appropriate check would be:

```swift
modifiers == [.option]
```

BashkaMen: fixed

Sources/KeyboardShortcuts/HotKey.swift
BashkaMen: I believe your application is more important and should run globally.
I removed all the hotkeys I found while setting up the workspace.

Sources/KeyboardShortcuts/HotKey.swift
```diff
@@ -324,6 +327,36 @@ final class HotKeyCenter {
...
+	private var hasOptionOnlyShortcuts: Bool {
+		hotKeys.values.compactMap(\.value).contains { hotKey in
+			let modifiers = NSEvent.ModifierFlags(carbon: hotKey.carbonModifiers)
+			return modifiers == [.option]
+		}
+	}
+
+	private func setGlobalOptionKeyMonitorEnabled(_ isEnabled: Bool) {
+		if isEnabled, globalEventMonitor == nil {
+			globalEventMonitor = NSEvent.addGlobalMonitorForEvents(matching: [.keyDown, .keyUp]) { [weak self] event in
```
sindresorhus, line 346: This requires accessibility access, so it's a no-go. From docs:

> Key-related events may only be monitored if accessibility is enabled or if your application is trusted for accessibility access (see [AXIsProcessTrusted()](dash-apple-api://load?request_key=ls/documentation/applicationservices/1460720-axisprocesstrusted)).

wojciech-kulik: Many apps use accessibility API. How about making it optional? Clients could decide wheter they want to use this API or not.

sindresorhus:
Thanks for the detailed investigation. However, this is not something I will accept.

`NSEvent.addGlobalMonitorForEvents` for key events requires Accessibility permission, and this package intentionally requires none. Without it the monitor silently does nothing, so for nearly all apps using KeyboardShortcuts this would have no effect — and it would intermittently double-fire whenever Carbon does deliver the event (a global monitor is passive, it can't replace the Carbon handler).

But more importantly, this is an intentional macOS restriction to curb key-logging of password characters (`⇧⌥O` → `Ø`), and there's no permission-free public API to reliably intercept Option + letter before the text input layer. I don't think this package should try to work around that.

